PR #4, opened 2015-08-12
Add `elm-test init`

avh4:
Make it easier to get started with testing by adding `elm-test init`, which does the following:

1. `elm package install deadfoxygrandpa/Elm-Test`
1. `elm package install maxsnew/IO`
1. Creates `TestRunner.elm` and `Tests.elm` with example tests.

### DIFF
--- a/bin/elm-test
+++ b/bin/elm-test
@@ -6,11 +6,29 @@ process.title = 'elm-test';
 
 var compile    = require("node-elm-compiler").compile,
   suffix       = require("../elm-io-ports.js"),
-  fs           = require("fs"),
+  fs           = require("fs-extra"),
+  path         = require("path"),
   temp         = require("temp").track(), // Automatically cleans up temp files.
   util         = require("util"),
   _            = require("lodash"),
   childProcess = require("child_process");
+
+if (process.argv[2] == "init") {
+  var copyTemplate = function(templateName) {
+    if (fs.existsSync(templateName)) {
+      console.log(templateName + " already exists");
+    } else {
+      fs.copySync(path.resolve(__dirname, "../templates/" + templateName), templateName);
+      console.log("Created " + templateName);
+    }
+  };
+
+  childProcess.execSync("elm package install deadfoxygrandpa/Elm-Test", {stdio:[0,1,2]});
+  childProcess.execSync("elm package install maxsnew/IO", {stdio:[0,1,2]});
+  copyTemplate("TestRunner.elm");
+  copyTemplate("Tests.elm");
+  process.exit(0);
+}
 
 var testFile   = process.argv[2],
     cwd        = __dirname;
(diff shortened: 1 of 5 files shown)


bin/elm-test
```diff
@@ -6,12 +6,30 @@ process.title = 'elm-test';
...
+    if (fs.existsSync(templateName)) {
+      console.log(templateName + " already exists");
+    } else {
+      fs.copySync(path.resolve(__dirname, "../templates/" + templateName), templateName);
+      console.log("Created " + templateName);
+    }
+  };
+
+  childProcess.execSync("elm package install deadfoxygrandpa/Elm-Test", {stdio:[0,1,2]});
+  childProcess.execSync("elm package install maxsnew/IO", {stdio:[0,1,2]});
```
rtfeldman: Probably want to include `--yes` on these, yeah?

templates/Tests.elm
```diff
@@ -0,0 +1,14 @@
...
+
+import String
+
+
+all : Test
+all = suite "A Test Suite"
+        [ test "Addition" (assertEqual (3 + 7) 10)
+        , test "String.left" (assertEqual "a" (String.left 1 "abcdefg"))
+        , test "This test should fail" (assert False)
+        ]
```
rtfeldman: I'd format this like so:

```elm
all : Test
all =
    suite "A Test Suite"
        [
            test "Addition" (assertEqual (3 + 7) 10),
            test "String.left" (assertEqual "a" (String.left 1 "abcdefg")),
            test "This test should fail" (assert False)
        ]
```

templates/Tests.elm
```diff
@@ -0,0 +1,14 @@
+module Tests where
+
+import ElmTest.Assertion as A exposing (assertEqual, assert)
```
rtfeldman: Seems reasonable to expose `assert` and `assertEqual` by default, but I think `as A` is more of a personal preference, and might surprise people.

package.json
```diff
@@ -28,6 +28,7 @@
   },
   "homepage": "https://github.com/rtfeldman/node-elm-test#readme",
   "dependencies": {
+    "fs-extra": "^0.23.1",
```
rtfeldman: Any objection to pinning this to exactly `0.23.1`?

rtfeldman:
This is awesome! Thanks for doing it. :smile_cat: 

I left a couple of comments.

bin/elm-test
avh4: I'm hesitant about that since people using `elm-test` for the first time might want to get the prompt for the new packages.  But how about if we add a `--yes` option to `elm-test init` ?

templates/Tests.elm
avh4: Okay (I had just copied this code from the examples folder)

templates/Tests.elm
avh4: Sounds good to remove `as A`

package.json
avh4: Sounds good.

avh4: The correct thing appears to be `~0.23.1`, which will match 0.23.*  http://stackoverflow.com/questions/22343224/difference-between-tilde-and-caret-in-package-json

avh4:
Made the suggested changes.

rtfeldman:
This rules!

PR #11, opened 2015-10-17
Fix for Elm 0.15, and fix CI

avh4:
@rtfeldman branch is not ready yet, but FYI I'm playing around with getting CI fixed

Hopefully can do a matrix build with 0.15.1 and 0.16

Unfortunately this will make CI really slow, but maybe next we can get someone to make an official apt-get package for elm, and then Travis can use that.


avh4:
Waiting for https://github.com/laszlopandy/elm-console/pull/4 to hopefully be merged so that solving this will be a bit easier.

avh4:
@rtfeldman this passes, and fixes 0.15!

avh4:
@rtfeldman thoughts on this?

PR #27, opened 2016-01-05
`elm-test` with no arguments runs tests/TestRunner.elm by default

avh4:
(Depends on #25 and #26)

This makes two changes:
  - running `elm-test tests/SomeFile.elm` where the filename begins with `tests/` will cause elm-test to cd into that directory before running the tests.  (Currently only for `tests/`, not for other directories).
  - running `elm-test` with no arguments is equivalent to running `elm-test tests/TestRunner.elm`

This means that getting started is easier!

```bash
mkdir my-project && cd my-project
elm-test init
elm-test
## You have test results!
```


rtfeldman:
I don't know of any other testing tools doing anything similar, but I'm on board!

Issue #22, opened 2015-12-22
Fix AppVeyor build


rtfeldman:
`npm test` runs `ci.sh`, which is a Bash script that works great on OS X and Linux (via Travis), but [which segfaults on Windows](https://ci.appveyor.com/project/rtfeldman/node-elm-test/build/1.0.17/job/dbx8no8jcdb2n6t6#L172) (via AppVeyor).

It would be great to port ci.sh to Node and run it via `npm test` instead of using a Bash script, so we can confirm that things are working on Windows.

PR #34, opened 2016-03-14
Port ci.sh to node

fapian:
Hey,

this PR should fix issue #22. Instead of a bash script node is used for the continuous integration script.
I used _shelljs_ to implement that script.

During the development I encountered an error. See lines 199 and 200.

https://ci.appveyor.com/project/fapian/node-elm-test/build/1.0.4/job/i3ul7613dh1xk89n#L199

The error originated from the _spawnElmPackageSync_ function. While trying to understand the error, 
I recognized that both calls of the method seem to be unnecessary, because the template _elm-package.json_ has both packages listed as dependencies.
Both packages will be installed, as soon as the CI script calls _elm-package install --yes_ and that's right after the _elm-test init_ function has finished.

So I removed both calls. As their are no more calls to that function, I removed the function as well.

If I can do anything...

Greetings
Fabi


rtfeldman:
Wonderful, thank you @fapian! 😻

PR #35, opened 2016-04-03
README change: path to TestRunner.elm

jessitron:
this is created in tests directory, gotta tell it that


rtfeldman:
😅 thanks @jessitron!

PR #37, opened 2016-04-09
Add FormData stub

lukewestby:
This creates a mostly-functioning `FormData` implementation that is good enough for testing things that call `Http.multipart`. Having it would be most helpful for testing `elm-http-extra` :smile: 

Thanks!


rtfeldman:
Seems reasonable, thanks!

Issue #38, opened 2016-04-17
Pass --warn flag to the Elm compiler


gkubisa:
I would like `node-elm-test` to support a `--warn` flag, which it would pass on to [node-elm-compiler](https://github.com/rtfeldman/node-elm-compiler). It would help to improve the quality of the test code.

I would also suggest using a library like [minimist](https://github.com/substack/minimist) or [yargs](https://github.com/yargs/yargs) for parsing the command line params in order to make the interface more flexible for the users and new flags easier to implement.

PR #55, opened 2016-08-12
Add `module` and `warn` arguments

### LINKED ISSUE #54: Support module names other than Main
Because of the way that we're structuring our code alongside other components, we would prefer not to use the nested `elm-package.json` style that elm-test uses by default.  It would be great if the elm-test runner was able to run tests from modules with other names.

bdukes:
This updates the command line parsing to use minimist (per the recommendation on #38).  It then adds a pass-through for `--warn`, and adds the ability to pass the module name of the test module.

This fixes #54, as well as fixing #38

### DIFF
--- a/bin/elm-test
+++ b/bin/elm-test
@@ -12,7 +12,8 @@ var compile    = require("node-elm-compiler").compile,
   util         = require("util"),
   _            = require("lodash"),
   spawn        = require("cross-spawn"),
-  minimist     = require("minimist");
+  minimist     = require("minimist"),
+  firstline    = require("firstline");
 
 var elm = {
   'elm-package': 'elm-package'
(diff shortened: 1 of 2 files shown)


mgold:
Looks good, this replaces some really ugly and brittle argument parsing code. cc @rtfeldman

rtfeldman:
Wow, this code looks GREAT! Thank you so much @bdukes! 😻 

One API consideration: I think a nicer API than passing `--module` in is to infer the module based on the file you pass. Otherwise you'll be doing stuff like `elm-test --module Foo.Bar.Baz Foo/Bar/Baz.elm` which seems a big redundant, no?

There is [code in `node-elm-compiler`](https://github.com/rtfeldman/node-elm-compiler/blob/master/index.js#L73-L79) that reads the `module` declaration out of a `.elm` file. (I could publicly expose that logic from `node-elm-compiler` if you'd like to use it.) It adds a dependency on [`firstline`](https://github.com/rtfeldman/node-elm-compiler/blob/master/package.json#L29) for this, because `firstline` is the highest-perf way I'm aware of to read just the first line of a file, and that's where the `module` declaration has to go.

Thoughts?

bdukes:
@rtfeldman, yeah, that makes sense to me.  I'm fine to just copy that couple of lines from `node-elm-compiler`, unless you feel like it would be better to be shared.

rtfeldman:
Copying seems fine to me!

bdukes:
I've pushed a commit that removes the `--module` argument and uses `firstline` to find the module name

rtfeldman:
Fantastic! Thank you so much @bdukes! 😻

PR #60, opened 2016-08-17
Find elm-package.json automatically

### LINKED ISSUE #59: Find elm-package.json for tests automatically
Currently, the code runs in the current directory unless the test file is in a `tests/` directory.  In that case, it jumps into the `tests/` directory and runs the elm compiler there.

Rather than assuming a particular file structure, the code should search for `elm-package.json` to determine where the compilation root is.

bdukes:
I noticed that the code was hard-coded to look for `tests/` (which I also noticed because it didn't work on Windows because of the different slash), and thought that it would be better to find `elm-package.json` instead.

Fixes #59

### DIFF
--- a/bin/elm-test
+++ b/bin/elm-test
@@ -13,7 +13,8 @@ var compile    = require("node-elm-compiler").compile,
   _            = require("lodash"),
   spawn        = require("cross-spawn"),
   minimist     = require("minimist"),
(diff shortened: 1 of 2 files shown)


rtfeldman:
Nice! Thanks again @bdukes! 😸

Issue #44, opened 2016-08-09
Fix init tests


rtfeldman:
[this](https://github.com/rtfeldman/node-test-runner/blob/e566684f0bf9804bf8812ab0b88e0bc8d7bb6ffd/tests/ci.js#L59-L60) and [this](https://github.com/rtfeldman/node-test-runner/blob/e566684f0bf9804bf8812ab0b88e0bc8d7bb6ffd/tests/ci.js#L69-L70) fail in `ci.js`, but work if you run the same steps manually.

The proximal cause is that for some reason in `ci.js`, `elm-test init` generates a template file that is consistently missing an entire line of code. If I run `elm-test init` in my terminal, the line of code is not missing.

benansell:
I had a quick look at this and think the difference in behaviour is due to the differences in the regex engines being used when attempting a multiline search. The [shelljs](https://github.com/shelljs/shelljs#sedoptions-search_regex-replacement-file_array) version is using javascript replace.

To avoid trying to find a way that a multiline regex would work in the terminal and via shelljs, I reframed the problem as "make the failing test in the template pass". I did consider ignoring it but I didn't see concise way of achieving that with elm-test.

rtfeldman:
Fixed by @benansell in #63

PR #63, opened 2016-08-20
Init test fix

benansell:
Feel free to ignore the 1st commit as it only adds semicolons - no functional changes - it just made the editing more pleasant once the IDE was happy.

2nd commit - proposed fix for issue #44. This code changes the sed regex's to update the failing test to a passing test rather than trying to remove it completely which is troublesome as it is across multiple lines.

### DIFF
--- a/tests/ci.js
+++ b/tests/ci.js
@@ -1,81 +1,81 @@
 #!/usr/bin/env node
 
-require('shelljs/global')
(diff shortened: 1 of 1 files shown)


rtfeldman:
Awesome! Thank you so much @benansell! Excellent work hunting down that fix. 😻

Issue #64, opened 2016-08-20
Feature Request: Add example fuzz test to Test.elm template


rohanorton:
Hey,

It would be really nice to have an example fuzz test in the test file that is produced by init. I think it might encourage people to try out fuzz testing, or at least provide a quick code example for those of us too lazy to go and look up the docs!!

Thanks

PR #67, opened 2016-08-26
Adds example fuzz test

benansell:
Not sure if this is what you were looking for as a fix for #64 

Adds a copy of elm-test example fuzz from docs added to template with slight rewording.

Added commented out Debug.log statement as a quick way for people to see the types of values being tested.

### DIFF
--- a/templates/Tests.elm
+++ b/templates/Tests.elm
@@ -1,5 +1,6 @@
 module Tests exposing (..)
 
+import Fuzz exposing (string)
 import Test exposing (..)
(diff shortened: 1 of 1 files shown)


rtfeldman:
Thanks @benansell! Per @mgold's preference for avoiding the `List.reverse` example, I'm gonna close in favor of https://github.com/rtfeldman/node-test-runner/pull/75 - but seriously, thank you for taking the time!

benansell:
No problem

PR #74, opened 2016-09-09
Add support for --watch

### LINKED ISSUE #56: File watching
It would be nice to support interactive testing where the runner watches loaded files for changes and reruns the tests when the files change. I think this would be particularly nice if it was optimized for speed so that once you hit save in  the editor you could see your tests finished by the time you switch to the console. To achieve this it would make sense to build these optimizations:

1. Work out the minimal subset of tests that need to run on a change. Given the following project:

   Main.elm imports A.elm and B.elm. B.elm imports C.elm. Each of these has a corresponding test file that only imports the file it tests. When I make a change to B.elm, I will only need to run tests from BTests.elm and MainTests.elm, since only these can have possibly changed.

   For this to work you would need to get the import tree from the compiler, which I don't know if it supported. When the dependency tree changes, it would make sense to re-run the entire suite.

2. Only run perhaps 10 fuzz tests rather than a hundred. Since you will run tests on every save, they should get decent coverage. You could perhaps run 100 on the first run.

Of course #51 would fit very well into this proposal as well.

athaeryn:
This adds support for running tests in watch mode, where the process will watch for changes to the source files and rerun the tests.

Would fix #56.

### DIFF
--- a/bin/elm-test
+++ b/bin/elm-test
@@ -238,7 +238,7 @@ findUp('elm-package.json', { cwd: path.dirname(testFile) })
                 unlinkDir: 'removed',
             };
 
-            watcher.on('all', (event, filePath) => {
+            watcher.on('all', function (event, filePath) {
                 var relativePath = path.relative(elmRootDir, filePath);
                 var eventName = eventNameMap[event] || event;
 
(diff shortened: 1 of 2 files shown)


rtfeldman:
Awesome, thanks so much @athaeryn! 😻

PR #76, opened 2016-09-13
Use downward arrow in error messages.

mgold:
Chalk.elm hard-coded the x icon and never printed the arrow. This fixes that.

(Unless you wanted to not use the arrow for some reason?)


rtfeldman:
Nope - thanks for the quick fix!

PR #80, opened 2016-10-01
Change format for Expect.equal to colored diff

jschomay:
Run bin/elm-test examples/tests/FailingTests.elm to see the colored diffs.

Uses string parsing to match and format only the Expect.equal tests, based
on how elm-test formats the messages.  This is non-invasive, but does
couple the parsing to that package.

### DIFF
--- a/bin/elm-test
+++ b/bin/elm-test
@@ -16,6 +16,8 @@ var compile    = require("node-elm-compiler").compile,
   firstline    = require("firstline"),
   findUp       = require("find-up"),
   chokidar     = require("chokidar");
+  jsdiff       = require('diff');
+
 
 var elm = {
   'elm-package': 'elm-package'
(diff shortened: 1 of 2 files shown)


jschomay:
This is one of the more "hacky-er" options from a few possible approaches.  See https://github.com/jschomay/elm-test/commit/a0d8bb655fc31776d9c6143acce4bc4c0c5144cf for a non-color approach.

In my opinion, the best option would be to remove the failure message formatting from elm-test and allow each test runner (html and node currently) to format failures appropriately.  This would require changing the Expectation type for Fail to contain the raw data needed to generate a message, instead of the message String.

I didn't take that approach because it is more invasive and requires PRs in 3 repos, but it may be better - more robust and cohesive.

jschomay:
Oooh, just saw that it failed...
I'm not sure why.  I was getting that same failure when I forked.

tracker-common:
We started using this on our own code, and unfortunately ran into a less than ideal experience for record diffs.  See the screenshot.  

Perhaps word-diff instead of letter-diff would be better!  Also, maybe formatting objects and arrays for output with newlines in appropriate places would help (and maybe line-diffs would be good enough then?)

screen shot 2016-10-13 at 2 31 42 pm

rtfeldman:
Ended up doing it like this in `0.18.11` for high-contrast accessibility (see #229). Thanks! ❤️

screen shot 2017-12-03 at 2 57 43 pm

PR #83, opened 2016-11-07
Fix tests/ using local node-test-runner for tests

stoeffel:
> Fix tests/ using local node-test-runner for tests

Currently master fails with:

```log
elm-package: elm-stuff/exact-dependencies.json: openBinaryFile: does not exist (No such file or directory)
elm-make: elm-stuff/exact-dependencies.json: openBinaryFile: does not exist (No such file or directory)
Compilation failed for Main.elm
elm-package: elm-stuff/exact-dependencies.json: openBinaryFile: does not exist (No such file or directory)
elm-make: elm-stuff/exact-dependencies.json: openBinaryFile: does not exist (No such file or directory)
Compilation failed for Main.elm
ci.js: ERROR: Expected tests to pass
npm ERR! Test failed.  See above for more details.
```

## What Changed ✏️ 

* adds tmp to gitignore
* fixes dependencies of templates
* use local node-test-runner for ci tests

PR #84, opened 2016-11-07
Fix polyfilling

stoeffel:
> Polyfill before creating `Elm` eval functions.

Some modules try to access `window` on `import`. i.e. https://github.com/elm-lang/animation-frame/blob/1.0.0/src/Native/AnimationFrame.js

## What Changed ✏️ 

* adds `Native.Polyfilled` to demo bug.
* moves polyfilling before `var Elm = function(module) { eval(compiledCode); return module.exports; }({});`

### Note 📝 

This includes commits from #83. Please merge that PR first and I will rebase.


rtfeldman:
Lovely, thank you! ❤️

PR #95, opened 2016-11-27
make sure to catch any error throw by evalElm

eeue56:
Currently if you cause a runtime error with Elm test on startup, you will get no output. This is because the promise is greedy and swallows it, meaning it will return that it successfully compiled the output, but nothing else. This leads to builds seeing the process return as though they passed when really they were causing runtime errors!


rtfeldman:
Good call!

PR #97, opened 2016-12-06
adds --doctest

stoeffel:
This allows a user to run doctests using `elm-test --doctest`. It will generate the doctests using https://github.com/stoeffel/elm-doc-test and then run them using `node-test-runner`.

Let me know what I should address.

Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ build/Release
 node_modules
 elm-stuff
 tmp
+examples/tests/Doc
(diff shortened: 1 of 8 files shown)


stoeffel:
@rtfeldman elm-doc-test uses template strings, which aren't supported in node 0.12. I actually think it's not necessary to still run ci on 0.12 size that is really old and not supported anymore.

rtfeldman:
Agreed, and thanks!

PR #98, opened 2016-12-07
Provide faster compile time on travis

amitaibu:
Apart of caching directories, we also use `libsysconfcpus` to compile it quicker. So elm-test runs on a warmed up cache.

p.s. Thanks to @eeue56 for the help.


rtfeldman:
Lovely, thank you!

PR #101, opened 2017-01-06
Also watch src dir in watch mode

robertjlooby:
@megapctr [pointed out](https://github.com/elm-community/elm-test/issues/104) that `elm-test --watch` doesn't watch the src dir. This fixes that.

### DIFF
--- a/bin/elm-test
+++ b/bin/elm-test
@@ -274,7 +274,8 @@ function runElmTest() {
           if (args.watch) {
               infoLog('Running in watch mode');
 
-              var watcher = chokidar.watch('**/*.elm', { ignoreInitial: true });
+              var watchedPaths = JSON.parse(fs.readFileSync(elmPackagePath, 'utf8'))['source-directories'].map(function(path) { return path + '/**/*.elm'; });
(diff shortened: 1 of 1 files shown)


megapctr:
Wow, that was fast! In the meantime, had just discovered that `findUp` returns the elm-package.json in the tests directory, which is what caused the original issue
```
var elmRootDir = path.dirname(elmPackagePath)
```
Perhaps changing this variable would be a better fix?

robertjlooby:
I think `elmRootDir` is correct. It needs to be the dir of the test app's `elm-package.json` since it is used as the working dir to build the test app.

megapctr:
oh, okay. In that case I would suggest renaming it to `elmTestDir` :)

avh4:
I think the correct fix here is to build the watch list from the `source-directories` list in `tests/elm-package.json`.  But I also think this PR is an improvement over the current behavior.

robertjlooby:
💯 @avh4 I hadn't even thought of that! Amended this PR to do that.

rtfeldman:
Thanks @robertjlooby!

PR #102, opened 2017-01-12
Add JUnit reporter for CI integration

### LINKED ISSUE #46: Output failures to JUnit XML
We should be able to output failures in a JUnit XML format, so CI servers can read it.

Notes:
- Test results should include duration, so we'll need to measure that
- It appears [Jenkins now supports nested `<testsuite>`s](https://issues.jenkins-ci.org/browse/JENKINS-6545), even though I believe they are not officially supported in the JUnit XML spec.

petejohanson:
Should fix #46.

Notes: I don't currently do deeply nested <testsuite> elements based on label. It's not clear to me that would really be beneficial. If you'd like me to investigate further, I'm happy to. Sample display a local Jenkins setup:

screenshot from 2017-01-12 16-26-13

### DIFF
--- a/bin/elm-test
+++ b/bin/elm-test
@@ -7,6 +7,7 @@ process.title = processTitle;
 var compile    = require("node-elm-compiler").compile,
   fs           = require("fs-extra"),
   chalk        = require("chalk"),
+  builder      = require("xmlbuilder"),
   path         = require("path"),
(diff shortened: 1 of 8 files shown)


rtfeldman:
This is awesome! Thank you @petejohanson!

Issue #104, opened 2017-01-25
Support TODO tests


mgold:
We've added `Test.todo` with the intent of the runners filtering out such tests, unless they are the only failing tests. Well, now that `Test.Runner.isTodo` exists it's time to implement that. The only trouble is that these changes are on master and not yet released, but we want the runners to be ready to go when we do release them.

Issue #100, opened 2017-01-06
watch mode should not get triggered by package downloads


avh4:
To recreate, use a project that has many dependencies listed in `tests/elm-package.json`:

```sh
rm -Rf tests/elm-stuff
elm-test --watch
```

On the first run, elm-test will download the dependencies, but when the dependencies are downloaded, watch mode will see a bunch of new elm files appear under elm-stuff, and will kick off a bunch of simultaneous new builds while the first build is still downloading dependencies (and all the new builds will start downloading dependencies too!)

I think there are two things that ideally should both be done:
  - Don't start a new build while a previous one is still running
  - Don't watch files under elm-stuff (though maybe some people would actually want this?)

rtfeldman:
> Don't watch files under elm-stuff

This part is implemented on a branch and will be in the next release.

I agree that the other one is a good idea, but it's harder. Might be worth a separate issue?

Issue #77, opened 2016-09-15
set default number of fuzz tests from command line


zkessin:
It would be nice to be able to run a test run where each fuzz test did a much larger number of tests than the normal without having to change code. 

that is being able to do this

```
elm test --fuzz-itterations 10000
```

Or via an environment variable, but probably better on the command line

rtfeldman:
What's a specific use case this would make nicer?

zkessin:
Being able to run more tests to find the obscure bug that might only happen 1 time in 10,000 or so. I have seen bugs that showup via a property based test (in erlang) that just required an unusual combination of factors, so being able to have a build that could run everything a larger number of times would be helpful.

rtfeldman:
Gotcha, makes sense. This is now implemented on a branch, and will be in the next release!

Issue #72, opened 2016-09-06
Tests in different path than tests/Main.elm


sporto:
In my project using previous version of elm-test I had 

`test/unit/`
`test/integration`

I was using `elm-make TestRunner.elm --output tests.js && node tests.js` to run the unit test.

I'm trying to upgrade to elm-test 2 and  `node-test-runner`. But it seems I need to put my test runner in `tests/Main.elm`.

I would like to put my Main file in `tests/unit/Main.elm`. Is there an option for this?

jinjor:
Have you tried `elm-test tests/unit/Main.elm`?

I just found this option from source code, but there is '--help' option too. I think these should be documented in README.

rtfeldman:
Update: the next release should make this nicer, because it will let you run `elm-test test/**/*.elm`

PR #107, opened 2017-02-13
Main-less

rtfeldman:
## What Changed

* Now you no longer write `Main.elm`. Instead, you pass whatever files you want to test, e.g. `elm-test tests/**/*.elm` and it runs just those files. (Fixes https://github.com/rtfeldman/node-test-runner/issues/72) Also it ignores paths in `elm-stuff`, which fixes https://github.com/rtfeldman/node-test-runner/issues/100
* Right now on this WIP branch, it just assumes each of those files will expose `suite : Test`, and fails if any of them do not.
  * This means if you try to put any test helpers in `elm-test tests/**/*.elm` such as something like `Fixtures.elm` and it doesn't expose `suite : Test`, the test run will fail.
  * That's a bummer.
  * A workaround would be putting all your helpers in a `test-helpers/` directory (sibling to `tests/`) and then putting `../test-helpers` in your `tests/elm-package.json`. Or specify 
  * This works, but is not great. A much better UX is if we incorporate https://github.com/stoeffel/elm-reflection - because it can actually (very quickly!) identify just those files that expose `suite : Test` and filter out the ones that don't.
* Added `--fuzz` CLI argument which overrides the default fuzz count of 100 (e.g. `--fuzz 42` would change it to 42. Fixes https://github.com/rtfeldman/node-test-runner/issues/77)
* Added `Test.todo` support (fixes https://github.com/rtfeldman/node-test-runner/issues/104)
* Added `--include` and `--exclude`, both of which take regular expressions to run on test descriptions, in case you want to filter that way instead of/in addition to by filename. (You can use both `--include` and `--exclude`; first it narrows down to only the `include` files, and then runs the `exclude` regex on those remaining ones.)
* Added `--add-dependencies` to add missing dependencies from one `elm-package.json` to another. (Fixes https://github.com/rtfeldman/node-test-runner/issues/28) Also, `elm-test init` does this now, which fixes https://github.com/rtfeldman/node-test-runner/issues/68
* If a test throws a runtime exception, that test now just fails and reports the exception message. (Fixes https://github.com/rtfeldman/node-test-runner/issues/69)


elm-package.json
```diff
@@ -1,19 +1,21 @@
...
-    "repository": "https://github.com/rtfeldman/node-test-runner.git",
+    "repository": "https://github.com/elm-community/elm-test.git",
     "license": "BSD-3-Clause",
     "source-directories": [
-        "src"
-    ],
-    "exposed-modules": [
-        "Test.Runner.Node"
+        "src",
+        "/Users/rtfeldman/code/elm-test/src/",
```
stoeffel: I guess we can remove this right?

example/tests/elm-package.json
```diff
@@ -0,0 +1,20 @@
+{
+    "version": "1.0.0",
+    "summary": "tests for node-test-runner example",
+    "repository": "https://github.com/elm-community/elm-test.git",
+    "license": "BSD-3-Clause",
+    "source-directories": [
+        "../src",
+        "/Users/rtfeldman/code/elm-test/src/",
```
stoeffel: same as above

package.json
```diff
@@ -31,17 +31,16 @@
...
     "firstline": "^1.2.0",
     "fs-extra": "0.30.0",
     "lodash": "4.13.1",
     "minimist": "^1.2.0",
     "node-elm-compiler": "4.1.0",
-    "temp": "0.8.3",
     "xmlbuilder": "^8.2.2"
   },
   "devDependencies": {
+    "glob": "^7.1.1",
```
stoeffel: :bug: this should be a dependency

PR #117, opened 2017-04-18
Hoist infoLog function up to the top level to fix broken references

tgecho:
I'm getting this exception when I run `--watch` mode:
```
ReferenceError: infoLog is not defined
    at generateAndRunTests (/Users/erik/code/node-test-runner/bin/elm-test:432:5)
    at /Users/erik/code/node-test-runner/bin/elm-test:261:9
```

It looks like this function was originally at the top level but got wrapped up in another function at some point. It seemed like it was kind of floating alone, so I moved it down next to where the flag is handled.

### DIFF
--- a/bin/elm-test
+++ b/bin/elm-test
@@ -294,11 +294,6 @@ function compileAllTests(testFilePaths) {
 }
 
 function generatePackageJson(filePathArgs) {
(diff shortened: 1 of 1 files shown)


rtfeldman:
Ha, thanks @tgecho!

PR #121, opened 2017-04-26
Ensure that an error is raised for invalid test modules

tgecho:
If multiple files are being tested and a valid one is checked before an invalid one is found, this [`resolve` call](https://github.com/rtfeldman/node-test-runner/pull/120/files#diff-b3b53682a18f203ac8d29b0e277cad26R371) causes the overall promise to succeed.

Simply removing the `resolve(line)` seems reasonably effective, but also seems to kind of subvert good promise hygiene.


rtfeldman:
I'm cool with that. Thanks @tgecho!

PR #133, opened 2017-05-12
give exit code 1 when there are unhandled exceptions

strickinato:
# What this does

Just like for all of the other error cases in the test runner, we should give a failure code when the tests don't pass. This adds `process.exit(1)` to uncaughtException branches

# How we got bit

In Jenkins pipeline we're looking out for exit codes to break the build. In our particular case, hitting Debug.crash was masking other real test failures because elm-test was dying but reporting exit code 0 which caused CI to secretly pass. It should die and tell everyone it dies.


rtfeldman:
Ah, bummer! These actually get translated into individual test failures in the next release, but still, this is a good idea in case something slips through the cracks.

Thanks @strickinato!

PR #144, opened 2017-05-25
Don't filter out port modules during verification

### LINKED ISSUE #143: Port module is identified as invalid module declaration in 0.18.3
When I migrated from the old elm-test version (0.18.2)  to the newer (0.18.3) by following the instructions [here](https://groups.google.com/forum/#!topic/elm-dev/7mmElgi4QM8) I ran into the following message:

```
$PWD/tests/Main.elm has an invalid module declaration. Check the first line of the file and make sure it has a valid module declaration there!
```

This was due to the fact that I left the `port` keyword in front of the `Main` module:

```
port module Main exposing (..)
```

Removing this fixed the issue. It may be nice to detect the left over `port` keyword and inform people that are migrating that they should remove this. The current error message is a bit misleading, because the declaration is actually valid and did not touch it in the migration process (before I removed the keyword).

zwilias:
Fixes #143

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -277,7 +277,7 @@ function verifyModules(filePaths) {
   return Promise.all(
     _.map(filePaths, function(filePath) {
       return firstline(filePath).then(function(line) {
-        matches = line.match(/^(?:port )?module\s+(\S+)\s*/);
+        matches = line.match(/^(?:(?:port|effect)\s+)?module\s+(\S+)\s*/);
 
         if (matches) {
           var moduleName = matches[1];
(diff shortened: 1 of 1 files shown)


gampleman:
Should a similar thing be there for effect modules?

zwilias:
I'm not sure to what degree an effect module exposing tests is a case we elm-test supports. On the other hand, the language supports it, and the tests-extractor supports it. So might as well :)

avh4:
The tests for the elm-testable rewrite have a dummy effect manager that is use used for some of the tests, so I need support for effect modules.  (Though I don't have tests in that file; but currently its existence prevents the other tests from running because of `tests/Test/EffectManager.elm has an invalid module declaration. Check the first line of the file and make sure it has a valid module declaration there!`.)

rtfeldman:
Thanks @zwilias!

PR #152, opened 2017-06-04
--report is broken

zwilias:
- [x] pass correct parameter to the Elm runner
- [x] Clean up redundant arg parsing code on the Elm side of things


rtfeldman:
@zwilias I'm inclined to release this bugfix sooner rather than later, since without it the entirety of `--report` is useless...do you mind if I merge as-is and then split the other checkboxes off as a separate PR after?

zwilias:
Sure, that's fine :)

zwilias:
Rebased on master so checks are green, continued remaining, more general tasks in #155

PR #153, opened 2017-06-04
Bug/fix travis

zwilias:
The _core_ thing is using `echo -e` when generating the sysconfcpu-ified `elm-make` - the file had literal `\n`'s in there, rather than actual newlines.

Other than that, I also switched travis to using the node-js specific containers so the whole nvm thing can be handled by travis itself, as well as using the `node_js` key to specify which node versions we should support.

Since mocha already executes `ci.js`, I removed the explicit `node ci.js` from the test-script, too.

It seems a bit odd to support `4.x` and (at the time of writing) `8.x`, though - perhaps we should consider something like 6, 7 and current to cover all bases?


rtfeldman:
😅 wooooow I cannot *believe* that was it! Thank you so much for finding this @zwilias!

As for 6 and 7, I'd rather not have Travis builds taking forever due to running a ton of versions. I'm open to revisiting if people file bugs for intermediate versions though!

Issue #112, opened 2017-03-31
Add documentation for `--report` flag


franklsf95:
This is minor but it took me a while to dig through the code to find out you can pass along `--report json` to generate JSON output instead of the default colorful command line output. It will be nice to include that in the documentation.

zwilias:
Covered in #155 🎉

Issue #86, opened 2016-11-09
Document cli configuration flags


halfzebra:
I think it would be super helpful, if [Configuration](https://github.com/rtfeldman/node-test-runner#configuration) section of REAMDE.md mentioned all available configuration options.

- [ ] yes
- [ ] warn
- [ ] version
- [ ] help
- [x] watch
- [x] compiler
- [x] seed
- [ ] report

zwilias:
#155 covers this :)

PR #155, opened 2017-06-05
document and test flags

zwilias:
(continuation of #152)

- [x] Add sanity tests for the different cli flags
  - [x] `--help`
    - [x] prints helps
    - [x] indicate failure so accidentally using it in CI will fail
  - [x] `--add-dependencies`
    - [x] add missing dependencies to the destination
    - [x] fail if destination does not exist
    - [x] fail if source does not exist
  - [x] `--compiler`
    - [x] runner should fail is argument cannot be executed
  - [x] `--seed`
    - [x] uses the provided seed - verified by checking the json output
  - [x] `--fuzz`
    - [x] defaults to 100
    - [x] uses the provided number
  - [x] `--report`
    - [x] `json` reports valid JSON lines (and only JSON lines)
    - [x] `junit` reports valid XML (and only XML)
  - [x] `--watch`
    - [x] tests should re-execute if the watched test is changed
- [x] Add info about the different cli flags to the readme (ref: #112, #86 )


zwilias:
`--report=json` and `--report=junit` currently also spits out the stdout of `node-elm-compiler` -> adding a concept of "machine readable" output formats, and passing options to the compiler process to only output stderr for those.

Furthermore, `--report=junit` fails: `ReferenceError: builder is not defined`

rtfeldman:
This looks great! Reading it made me notice a couple things, but I'll make a PR rather than respond.

Thanks so much @zwilias, this is really fantastic! 😻

PR #161, opened 2017-06-14
Fix condition for reporter options

jfmengels:
While looking at the source code, I came across this line. The conditions always evaluated to true, regardless of the reporter passed to it, which I assume is not the intended behavior.


zwilias:
Whoosh, totally missed that! Good catch, thanks @jfmengels

Issue #165, opened 2017-06-15
Dot in github repository name crashes elm-test


antivanov:
When there is a dot in the repository name elm-test crashes with a cryptic error. Reproducible with the repository https://github.com/antivanov/underscore.elm

```
$ elm-test
Unhandled exception while running the tests: TypeError: Cannot read property '1' of null
    at Object.repositoryToNativePackageName (/home/anton/Apps/node-v6.11.0-linux-x64/lib/node_modules/elm-test/lib/runner.js:55:12)
    at generatePackageJson (/home/anton/Apps/node-v6.11.0-linux-x64/lib/node_modules/elm-test/bin/elm-test:408:12)
    at runElmTest (/home/anton/Apps/node-v6.11.0-linux-x64/lib/node_modules/elm-test/bin/elm-test:279:22)
    at Object.<anonymous> (/home/anton/Apps/node-v6.11.0-linux-x64/lib/node_modules/elm-test/bin/elm-test:562:1)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.runMain (module.js:604:10)
```

Looking into the source code or runner.js I found that the following regex is used to parse the repository name:

```js
  var matches = repository.match(/\/([^\/]+)\/([^\/\.]+)\.git/);
```

Looks like the regex should be modified and the last dot exclusion removed. Then the new regex will match all the previous package names and also the ones that include dots in them. I will create a PR.

zwilias:
If I'm not mistaken, elm-lang/elm-make#106 basically means you could get runtime errors because of this. Perhaps instead of allowing it, we should display a better error message?

antivanov:
Yes, @zwilias, good point. Actually allowing dot in the repository name at this point just leads to the elm-make failing later. At least we should then provide a better error message, I will update the pull request. Dots will not be allowed still, but this will be communicated clearer.

PR #166, opened 2017-06-15
Show understandable error message when repository name contains a dot in 'repositoryToNativePackageName'

antivanov:
Fixes https://github.com/rtfeldman/node-test-runner/issues/165

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -53,22 +53,18 @@ function repositoryToNativePackageName(repository/*:string*/) {
   var matches = repository.match(/\/([^\/]+)\/([^\/]+)\.git/);
 
   if (matches instanceof Array) {
-    var username = matches[1];
+    var userName = matches[1];
     var repoName = matches[2];
 
     //"A dot in repository name breaks `elm-make`" https://github.com/elm-lang/elm-make/issues/106
-    if (username.indexOf('.') >= 0 || repoName.indexOf('.') >= 0) {
+    if (userName.indexOf('.') >= 0 || repoName.indexOf('.') >= 0) {
       throw ("Elm currently doesn't support having periods in the user/project part of the repository field of elm-package.json. Aborting test run.");
     }
-    if (typeof username === "string" && typeof repoName === "string") {
-      // From the above example, return "rtfeldman$node_test_runner"
-      return [
-        matches[1].replace(/-/g, "_"),
-        matches[2].replace(/-/g, "_")
-      ].join("$");
-    } else {
-      throw ("Unable to convert repository " + repository + " to package name.")
-    }
+    // From the above example, return "rtfeldman$node_test_runner"
+    return [
+      userName,
+      repoName
+    ].map(name => name.replace(/-/g, "_")).join("$");
   } else {
     throw ("Unable to convert repository " + repository + " to package name.")
   }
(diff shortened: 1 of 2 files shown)


lib/runner.js
```diff
@@ -50,12 +50,19 @@ function repositoryToNativePackageName(repository/*:string*/) {
...
   // matches[2] = "node-test-runner"
-  var matches = repository.match(/\/([^\/]+)\/([^\/\.]+)\.git/);
+  var matches = repository.match(/\/([^\/]+)\/([^\/]+)\.git/);
 
   if (matches instanceof Array) {
     var username = matches[1];
     var repoName = matches[2];
 
+    //"A dot in repository name breaks `elm-make`" https://github.com/elm-lang/elm-make/issues/106
+    if (repoName.indexOf('.') >= 0) {
```
zwilias: While we're here, might as well check the username. Even though github doesn't allow periods in usernames, there is no requirement that the repo _actually exists_, so this can happen.

lib/runner.js
```diff
@@ -50,12 +50,19 @@ function repositoryToNativePackageName(repository/*:string*/) {
...
+      console.error("Repository name '" + repoName + "' includes the dot symbol, this can cause issues with 'elm-make'");
+      console.error("Please rename the repository or elm-test will not be able to run the tests.");
```
zwilias: Good error messages are hard. As an alternative, what do you think of something like this?

_Elm currently doesn't support having periods in the `user/project` part of the `repository` field of `elm-package.json`.
Aborting test run._

lib/runner.js
antivanov: Yes, good point, we should check the user name for dots as well.

lib/runner.js
antivanov: Yes, this sounds a bit clearer, will update the error message.

antivanov:
Will also try to add unit tests for the changes and will update the pull request.

lib/runner.js
antivanov: Now both user name and the repository name are checked for dots.

lib/runner.js
```diff
@@ -50,21 +50,21 @@ function repositoryToNativePackageName(repository/*:string*/) {
...
   // matches[2] = "node-test-runner"
-  var matches = repository.match(/\/([^\/]+)\/([^\/\.]+)\.git/);
+  var matches = repository.match(/\/([^\/]+)\/([^\/]+)\.git/);
 
   if (matches instanceof Array) {
-    var username = matches[1];
+    var userName = matches[1];
     var repoName = matches[2];
 
-    if (typeof username === "string" && typeof repoName === "string") {
```
antivanov: I checked that as we require the matching groups to match at least one symbol each at this point both `userName` and `repoName` are not empty strings. Regular expression guarantees that this check is always true if there is a match:

```javascript
"https://github.com/rtfeldman/node-test-runner.git".match(regex)
(3) ["/rtfeldman/node-test-runner.git", "rtfeldman", "node-test-runner", index: 18, input: "https://github.com/rtfeldman/node-test-runner.git"]
"https://github.com//node-test-runner.git".match(regex)
null
"https://github.com/rtfeldman/.git".match(regex)
null
"https://github.com/123/node-test-runner.git".match(regex)
(3) ["/123/node-test-runner.git", "123", "node-test-runner", index: 18, input: "https://github.com/123/node-test-runner.git"]
```

antivanov:
Updated the pull request, please, take a look when there is time.

zwilias:
Thanks for the effort, @antivanov!  🏆 

@rtfeldman wdyt?

rtfeldman:
Love it, thanks @antivanov! 😻

antivanov:
Thanks a lot :)

PR #196, opened 2017-08-04
silence installer when we run json/junit reporter

stoeffel:
If you pipe the output from `report junit` into a file then the xml is sometimes not valid. The reason for this is that `binstall` runs with `verbose: true`. So `
Downloading binaries from https://dl.bintray.com/elmlang/elm-test/0.18.8/linux-x64.tar.gz`
is in the produced xml. This PR attempts to run binstall only in verbose mode when elm-test runs in a none-machinereadable way.
this was caught by @ento ❤️

### DIFF
--- a/install.js
+++ b/install.js
@@ -1,7 +1,7 @@
 var path = require("path");
 var install = require(path.join(__dirname, "installer"));
 
-install().then(
+install(true).then(
   function(successMessage) {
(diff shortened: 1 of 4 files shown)


rtfeldman:
Lovely, thanks @stoeffel!

PR #197, opened 2017-08-04
Chokidar file watcher: Await write finish

### LINKED ISSUE #194: Cancel previous running compile process in watch mode
I've noticed `node-test-runner` started in watch mode is looking for file changes and triggers build and tests. This is not aware of currently running compile or test processes.

If you hit `save` several times, this is what will happen:
screenshot from 2017-08-02 18-52-24
screenshot from 2017-08-02 18-53-18

This is for demonstration but it happens for me while developing: Elm-format will run on save and triggers `elm-test` twice. If it's possible, I think it could be a solution to cancel currently executed tasks before starting a new compile process. Or is this a known issue or not possible to recognize?

Issue could be related to #110

andys8:
Change [chokidar](https://github.com/paulmillr/chokidar) settings to await file change. This fixes the problem that `elm-format` configured on save, will trigger `elm-test` in watch mode twice. It will lead to `500 ms` latency in watch mode, but it seems to be acceptable in comparison to running twice.

Fixes #194

### DIFF
--- a/lib/elm-test.js
+++ b/lib/elm-test.js
@@ -639,6 +639,9 @@ function generateAndRunTests(tests, filePathArgs, generatedSrc, getGlobs) {
     var watchedPaths = watchedSourcePaths.concat(watchedTestPaths);
 
     var watcher = chokidar.watch(watchedPaths, {
+      awaitWriteFinish: {
+        stabilityThreshold: 500
+      },
       ignoreInitial: true,
       ignored: /(\/|^)elm-stuff(\/|$)/
     });


andys8:
[Oh, oh ^^](https://ci.appveyor.com/project/rtfeldman/node-test-runner/build/1.0.485/job/iv4n9as1asp1ccfw)
Relevant test if failing with node v4, but not with v5. Needs to be evaluated if the test itself is vulnerable to timing issues or if there is a problem with this change an node v4.

andys8:
That's interesting. `nodejs 4.0.0` will not work, but it *sometimes* won't work with the `master` branch, too.
But there is a main difference:
Travis is using `v4.8.4`
image
AppVeyor is using `v4.0.0`
image

This is uncommon, right? Who is the target group still using the initial release node 4? I'd say this is not intended.

And with the latest node v4 version `v4.8.4` both branches are working fine: `rtfeldman:master` and `andys8:chokidar-watcher-settings`.

I'll open an additional pull request in case this inconsistent behavior is not wanted.

rtfeldman:
It's kinda flakey. I re-ran and it passed. 😛

rtfeldman:
Looks good to me! Thanks @andys8

PR #200, opened 2017-08-14
add --skip-install flag

jschomay:
I often code on the bus or places without internet.  Because of the missing packages check, I can't run my tests in these situations.

This change makes it possible to opt out of the package check so I can run my tests again without wifi.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,14 @@ source-directories), your tests will get rerun.
 elm-test --watch
 ```
 
+#### `--skip-install`
+
+Skips the missing packages check.  Useful for when offline.  Just make sure all of the necessary packages have previously been installed.
+
+```
+elm-test --skip-install
+```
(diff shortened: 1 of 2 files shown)


avh4:
Can we fix this in a way that doesn't add a new option?  Maybe one of:

 - warn and continue if the install fails
 - silently ignore the install failure if all the necessary dependencies exist
 - only try installing if the network is available
 - detect when elm-package fails because of the network (ideally it should return a different error code for that case?) and continue if that was the failure reason

rtfeldman:
@jschomay Thank you! This looks like a very lightweight approach.

rtfeldman:
@jschomay Published as `elm-test@beta` so you can use it right away. (I have some other stuff I want to get in before doing a full normal release.)

Thanks again!

Issue #201, opened 2017-08-16
elm-test hanging on circleci


97jaz:
(And yes, I'm running it with `sysconfcpus -n 2 elm-test`.)

I only have two tests at the moment. (Unfortunately, the code is not open source, or else I'd post them.)

The closest issue I found was #53, but I'm not running on Windows, and neither of the tests are fuzz tests. The output is:
```
$ #!/bin/bash -eo pipefail
sysconfcpus -n 2 elm-test

Success! Compiled 105 modules.
Successfully generated /dev/null
Success! Compiled 11 modules.
Successfully generated /home/circleci/connect/elm-stuff/generated-code/elm-community/elm-test/elmTestOutput.js

elm-test 0.18.8
---------------

Running 2 tests. To reproduce these results, run: elm-test --fuzz 100 --seed 745527669
```

And then it never completes. Circle CI allows you to ssh into the box to debug a build. When I did that and ran `elm-test` myself, it worked the first time and hung thereafter.

Any thoughts on how to debug this further?

97jaz:
Oh, and this is probably obvious, but it never hangs when I run it locally (on OS X). The CircleCI container is Ubuntu Trusty. The tests themselves, and the code they test, are deterministic. They should always terminate.

97jaz:
Okay, it appears that when this happens there are a ton of node processes running, which suggests that the effect of `sysconfcpus -n 2` isn't carried over to whatever subprocesses are created by `elm-test`.

... except I'm pretty sure it's supposed to be. I also tried:
```bash
export LIBSYSCONFCPUS=2
sysconfcpus  elm-test
```
which didn't help.

97jaz:
And... that's not surprising, since node's `os.cpus()` isn't affected by `sysconfcpus`.

So that's the issue: `sysconfcpus` just doesn't have any effect on how many processors `node` thinks there are: https://github.com/rtfeldman/node-test-runner/blob/master/lib/elm-test.js#L59

Maybe a `--cpus` or `--workers` flag is in order.

(That said, I don't know for a fact that this is what's causing `elm-test` to hang.)

97jaz:
I apologize for the many comments in succession, but I just tried with version 0.18.7, and it worked, so I do suspect that `os.cpus()` is the culprit here.

rtfeldman:
[...]

PR #203, opened 2017-08-18
Adds a "processes" flag to specify parallelism explicitly

97jaz:
@rtfeldman This change works for me as a solution to #201, but:
1. You may not want to add a flag.
2. Since `elm-test` invokes `elm-make` and the two programs use different mechanisms to choose how many jobs to use, I still need to use both `sysconfcpus` and this new `--processes` flag when running `elm-test`, which is clearly gross. I don't have a solution in mind for this.

Oh, the `fsevents` change: without this, my install fails when I'm building a Docker image for CircleCI:
```
error fsevents@1.1.2: The platform "linux" is incompatible with this module.
error Found incompatible module
```

### DIFF
--- a/lib/elm-test.js
+++ b/lib/elm-test.js
@@ -51,12 +51,15 @@ var args = minimist(process.argv.slice(2), {
     compiler: "c",
     "add-dependencies": "a",
     report: "r",
-    watch: "w"
+    watch: "w",
+    processes: "p"
   },
   boolean: ["warn", "version", "help", "watch"],
-  string: ["add-dependencies", "compiler", "seed", "report", "fuzz"]
+  string: ["add-dependencies", "compiler", "seed", "report", "fuzz", "processes"]
 });
-var processes = Math.max(1, os.cpus().length);
+
+var maxProcesses = parseInt(args.processes) || os.cpus().length;
+var processes = Math.max(1, maxProcesses);
 
 // Recursively search directories for *.elm files, excluding elm-stuff/
 function resolveFilePath(filename) {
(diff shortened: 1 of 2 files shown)


rtfeldman:
So I suspect we can auto-detect this. 

Let's assume `process.pid` is `1234`. In that case, we can exec this command:

```
$ taskset -p -c 1234
pid 1000's current affinity list: 0-15
```

Theoretically, that affinity list (`0-15` meaning "CPU core IDs 0 through 15", so 16 cores in total available, although apparently sometimes it can be a comma-separated list like `0,1` meaning "CPU core IDs 0 and 1") tells us how many processors are *actually available*.

Since I suspect virtualization will only come up on Linux systems, I think it's worth checking if `process.platform` is `"linux"` and if so, executing `taskset` (which should be present on any Linux system) and reading its `stdout` in order to detect how many processors are available.

However, before doing any of that, it would be wise to verify that `taskset` actually reports the number of cores we're looking for on CircleCI. 😅 

@97jaz would you be able to SSH in and see what `taskset` reports?

97jaz:
@rtfeldman A good thought, but it doesn't seem to help:
```
circleci@f94d59160c8b:~$ taskset -p -c 144
pid 144's current affinity list: 0-35
```

97jaz:
I just opened a support ticket with CircleCI to see if they know a reliable way to get the number of available cores from within a container.

rtfeldman:
Nice, thanks! Here's hoping. 🤞

97jaz:
CircleCI support was not helpful. The first response I received suggested reading the files in `/sys/fs/cgroup/cpuset/`, except those also report 36 cores. The second person confirmed that`/sys/fs/cgroup/cpuset/` does not give the correct information and confirmed that there are 2 cores available but did not answer the question of whether there's any way to determine that without knowing it *a priori*.

rtfeldman:
FYI, I found a way to make this Just Work (on Travis) with no flag and no hardcoding knowledge of Travis.

However, the fix is part of a larger effort I'm working on, and I don't think I'll have a beta ready for another week or two. 😅

harrysarson:
Given this has been two years and the linked issue is closed I will close!

PR #206, opened 2017-08-30
Highlighted Diffs

rtfeldman:
👉 **This has been published as `elm-test@beta` if anyone wants to try it. Feedback welcome!** 👈 

This is a noisy diff because I decided to inline `rtfeldman/console-print` and `jinjor/elm-diff` to avoid potential conflicts with `elm-test` users' installed packages.

The console runner now prints out colored diffs for `Expect.equal` like so:

screen shot 2017-08-29 at 11 54 20 pm

If colors are not supported (e.g. because you're on CI), it renders arrows instead:

screen shot 2017-08-30 at 12 11 13 am

Also, previously the JUnit and JSON reporters would output the same unicode art pipes as the console, which looked fairly ridiculous. 😅 

Now they output normal-looking messages.


zwilias:
This is beautiful. 😻

lib/elm-test.js
```diff
@@ -449,11 +450,22 @@ function generatePackageJson(filePathArgs) {
...
   if (!elmPackageContents.dependencies) {
     elmPackageContents.dependencies = {};
   }
 
+  if (
+    !elmPackageContents.dependencies.hasOwnProperty(
+      "Skinney/elm-array-exploration"
+    )
+  ) {
+    // Highlightable needs jinjor/elm-diff to render diffs; we have inlined jinjor/elm-diff
```
zwilias: So `jinjor/elm-diff` depends on `Array.Hamt`?

elm-package.json
```diff
@@ -1,18 +1,20 @@
...
+    "license": "BSD-3-Clause",
+    "source-directories": [
+        "src",
+        "native-src"
+    ],
+    "exposed-modules": [],
+    "native-modules": true,
+    "dependencies": {
+        "Skinney/elm-array-exploration": "2.0.5 <= v < 3.0.0",
+        "elm-community/elm-test": "4.2.0 <= v < 5.0.0",
```
zwilias: Hm, will this break for people who are already >=0.18.3 but still have elm-community/elm-test@4.0.x or 4.1.x? Is this something we could detect and offer to fix when running this version?

src/Diff.elm
```diff
@@ -0,0 +1,319 @@
...
+   FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+   DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+   SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+   CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+   OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+   OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+-}
+
+import Array.Hamt as Array exposing (Array)
```
zwilias: Does it make sense to use core's array? On principle, I agree with using `Array.Hamt` _but_ it adds an extra dependency, and requires extra code to ensure that dependency is there. I don't see unsafe usage (it's mostly just access by index and `Array.fromList`, rather than slicing and comparing), so I think that's feasible.

src/Test/Reporter/Highlightable.elm
```diff
@@ -0,0 +1,48 @@
...
+resolve { fromHighlighted, fromPlain } highlightable =
+    case highlightable of
+        Highlighted val ->
+            fromHighlighted val
+
+        Plain val ->
+            fromPlain val
+
+
+fromLists : List a -> List a -> List (Highlightable a)
```
zwilias: It doesn't seem obvious that `Highlightable.fromLists` will do a `diff`. Is there a better name for this?

rtfeldman: 👍 Renamed to`diffLists` in 5616fa615ad97f2a88341466115ccae67267b6bb

PR #214, opened 2017-09-10
Polyfill document.location

### LINKED ISSUE #213: Consider polyfilling document.location
A while ago I looked into [this StackOverflow question](https://stackoverflow.com/questions/44909872) where a user was hitting a JavaScript exception running elm-test.

It turns out that the user was importing what appears to be the main module of his app in his tests.  This module includes a call to `Navigation.program`, and the native code behind that accesses `document.location`. The test runner's polyfilled `document` object doesn't define a `location` property, hence the JS exception.  The tests ran if I commented out the call to `Navigation.program`.

I'm aware you have guidance against importing main modules in tests, but a JS exception such as the one this user saw is a bit of a gotcha.  It wasn't obvious to the user what his mistake was nor how to fix it.  I was only able to figure out what was going on by modifying lib/runner.js on my machine to spit out the JS code it generated.

The `document` polyfill appears to be set up in templates/before.js (it used to be in lib/runner.js) and is currently

    var document = { body: {}, createTextNode: function() {} };

To avoid problems such as this in the future, could it be changed to the following instead?

    var location = { href: "", host: "", hostname: "", protocol: "", origin: "", port: "", pathname: "", search: "", hash: "", username: "", password: "" };
    var document = { body: {}, createTextNode: function() {}, location: location };

I made this modification to lib/runner.js on my machine and it fixed the problem, in that I could run the tests with the call to `Navigation.program` reinstated.

rtfeldman:
Fixes #213

### DIFF
--- a/templates/before.js
+++ b/templates/before.js
@@ -5,7 +5,20 @@ var window = {
   removeEventListener: function() {}
 };
 
-var document = { body: {}, createTextNode: function() {} };
+var location = {
+  href: "",
+  host: "",
+  hostname: "",
+  protocol: "",
+  origin: "",
+  port: "",
+  pathname: "",
+  search: "",
+  hash: "",
+  username: "",
+  password: ""
+};
+var document = { body: {}, createTextNode: function() {}, location: location };
 
 if (typeof XMLHttpRequest === "undefined") {
   XMLHttpRequest = function() {


zwilias:
LGTM!

PR #221, opened 2017-09-29
Document that elm-test installs packages

jackfranklin:
I didn't realise that this was the case - see https://github.com/halfzebra/create-elm-app/issues/182 for the context.


zwilias:
Thanks for the contribution, @jackfranklin!

Issue #210, opened 2017-09-04
Always update tests/elm-package.json


boxed:
It would be nice if the "--add-dependencies" command was always run when doing "elm-test". Then point 4 from "Running tests locally" at https://github.com/elm-community/elm-test could could be removed and it would make the development experience smoother.

Alternatively the runner could error out if you haven't updated the dependencies and tell you to run the add dependencies command.

boxed:
I’m confused. Did you just close this issue because a command was renamed?

rtfeldman:
No. `--add-dependencies` is gone because it wouldn't have any effect in the Elm 0.19 version of `elm-test`, so the request to run it every time `elm-test` gets run wouldn't make any sense either.

boxed:
But then you wrote it has "been replaced with elm-test install"? So will elm-test install do what "add-dependencies" did before? If so then this issue should just have the title and description updated to match the new reality, not just being closed.

boxed:
@zwilias Thanks. Yes, that explains it. Thanks.

PR #222, opened 2017-10-15
Change default repository path to avoid elm-test #210

drathier:
The previous default is causing namespace collisions with the project we're testing, so we need this path to be different than the path that elm-test itself is using. 
https://github.com/elm-community/elm-test/issues/210


rtfeldman:
@drathier published as `elm-test@beta` - can you verify that it works now?

drathier:
Double-checked. It works. Thanks :)

PR #228, opened 2017-12-01
Split on newlines rather than reading as an array

zwilias:
A single `socket.write` does not always result in a single
`socket.on('data')` event being triggered. The way this has manifested
in the past has been by multiple messages being concatenated into a
single message. To that end, the current strategy is to join messages
into an array by appending `,` on the sending side and transforming that
on the receiving side.

However, a new, [reproducible case][repr] has popped up where the inverse is
happening: a single write resulting in 2 separate chunks of data coming
in. This isn't really something we can handle with the current strategy.

Turns out that this is essentially a solved problem: the `split` package
exports a `split()` function that a socket can be piped through and
which does all the bookkeeping to split chunks into proper messages.

So the new strategy is `append a newline`, and `split on newlines`. All
in all, this looks more like plain old json-lines.

[repr]: https://github.com/magopian/elm-scce

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -157,7 +157,7 @@ function findTests(
   testFilePaths /*: Array<string>*/,
   sourceDirs /*: Array<string>*/,
   verbose /*: boolean*/
-) {
+)/*:Promise<void>*/ {
   return new Promise(function(resolve, reject) {
     function finish() {
       var process = spawn(readElmiPath, ["--path", elmPackageJsonPath]);
--- a/lib/supervisor.js
+++ b/lib/supervisor.js
@@ -6,7 +6,8 @@ var chalk = require("chalk"), // Chalk is used only for the "Watching for change
   fs = require("fs-extra"),
   net = require("net"),
   child_process = require("child_process"),
-  pipeFilename = require("./pipe-filename.js");
+  pipeFilename = require("./pipe-filename.js"),
+  split = require("split");
 
 function run(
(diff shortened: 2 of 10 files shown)


rtfeldman:
What if the runner sends a failure message string which contains newlines?

zwilias:
`Encode.string` escapes newlines to a literal `\` followed by `n`, while the newline that is appended is a literal newline.

rtfeldman:
Got it, great work solving this! 🎉

One request: would you mind adding a test which reproduces that SSCCE, so we can confirm that it no longer breaks after this change?

zwilias:
Good point. In setting the SSCCE up, I noticed that the Elm tests in `tests/` aren't actually tested since we don't leave `example/` before grabbing tests from `tests/`.

On a separate note, the SSCCE only fails when color is supported, so I'm taking the easy way out and forcing all those tests to run with colorized output enabled by adding the [`--color` flag](https://www.npmjs.com/package/supports-color#info). I hope that's okay.

So, doing this in phases to ensure that (1) the failure manifests and (2) this PR fixes it:

- re-enable tests: https://travis-ci.org/rtfeldman/node-test-runner/builds/310800790
- add failing test (only fails on OSX): https://travis-ci.org/rtfeldman/node-test-runner/builds/310803149
- merge `reenable-tests` in this branch, should be all green again.

rtfeldman:
Boom, love it! Thank you @zwilias!

Issue #219, opened 2017-09-19
elm-test hangs


smucode:
I'm having issues with elm-test hanging, both on my 2016 Macbook Pro, a Dell XPS 15 and on our CI server.

On my MBP I'm consistently able to reproduce (in a fresh ubuntu vm) by:

1. Installing nvm and node 8
2. Installing elm and elm-test via npm
3. Running `elm-package install` and `elm-test init`
4. Running `elm-test` in a loop

Using this approach it appears to hang after ~25 runs on my MBP. Same results trying something similar on native OSX.

On the Dell XPS 15 the raciness is even worse, hangs about 1 in 4 runs.

I also did try using elm-test directly from master (with the unix-socket changes), and was a lot less racy, but it still hung from time to time.

Downgrading to `elm-test 0.18.7` solved the problem.

Let me know if there is anything else you need. Appreciate the great work!

isaacseymour:
I'm experiencing this too, but maybe slightly different:

- very very frequent hangs in CI (using CircleCI v2, tests run inside a docker container)
- yet to see a hang on my Macbook (but I haven't tried much either)

This occurred only when upgrading 0.18.7->0.18.9.

leifg:
I can confirm this. I also found out that after downgrading node to `8.2.1` I did not experience this error anymore (see my [comment](https://github.com/rtfeldman/node-test-runner/issues/201#issuecomment-328762394))

smucode:
I tried to reproduce on node 8.2.1 and can confirm it does hang after running elm-test in a loop for a minute or two (running the string reverse example test).  It does seem to happen less frequently than before, but that might be to me upgrading to high sierra or some other random factor.

giulioungaretti:
Same issue with the following set-up (running the example test from elm-test) :

OS: MacOS high Sierra
NODE: v8.5.0
elm-test 0.18.9

7er:
Same issue here. elm-test always hangs if there is a compile error in one of the test files:
OS: Ubuntu Linux 17.04
NODE: v8.7.0
elm-test: 0.18.6

Reverting to node v7.9.0 removes the problem

jasononeil:
We've encountered this problem too, but I'm fairly sure it was solved by this 4ecd1068f31e221104454b85d6c0b2c8b8f9902c (the very first commit after 0.18.9).

Setup to reproduce:

```shell
> cd node-test-runner/
> git checkout 432556e
[...]

PR #231, opened 2017-12-04
Allow running multiple instances of elm-test in parallel

zwilias:
Using the process ID of the main process in the name of the sockets
means multiple parallel instances of elm-test won't clash or try to
use the same socket across multiple processes.

Should take care of @smucode's issue in #219


smucode:
Yep, can confirm. Not able to reproduce my issue any more.

rtfeldman:
Fantastic! Thaks @zwilias! 😻

PR #233, opened 2017-12-07
Allow directories other than "test" or "tests"

liamcurry:
Given the following file structure:

```
src/
    Nested/
        Tests.elm
elm-package.json
```

I would expect to be able to run `elm-test src/Nested/Tests.elm` to run any tests in that file. However this currently isn't allowed because it isn't in the `test` or `tests` directory. Errors like this are thrown:

```sh
[1] % elm-test src/Nested/Tests.elm
It looks like you're running elm-test from within your tests directory.

Please run elm-test from your project's root directory, where its elm-package.json lives!
```

With the changes in this PR I was able to run tests successfully, but I have a feeling there's still some work to do with ignoring nested `node_modules` and `elm-stuff` directories.

### DIFF
--- a/lib/elm-test.js
+++ b/lib/elm-test.js
@@ -310,7 +310,7 @@ function runElmTest() {
     );
 
     getGlobs = function() {
-      return globifyWithRoot(root, "test?(s)/**/*.elm");
+      return globifyWithRoot(root, "**/*.elm");
     };
   }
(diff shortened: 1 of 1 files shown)


rtfeldman:
We ended up going the other direction - it's just `tests/` now. 😄

PR #235, opened 2017-12-08
Fix `--watch`

### LINKED ISSUE #234: `--watch` is broken
After executing the tests, it dumps some nonsense:

```
Unhandled exception while running the tests: SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at Stream.<anonymous> (*/node-test-runner/lib/supervisor.js:122:27)
    [...]
```

Thanks to @iamvery for surfacing this on Slack!

zwilias:
When a worker is done, the socket is `drain`ed.

This results in an extra albeit empty message being received.
Previously, the "make it an array" made this invisible, but now that we
split on newlines, we attempt to `JSON.parse` an empty
string.

So, easy fix: ignore empty messages.

Resolves #234

### DIFF
--- a/lib/supervisor.js
+++ b/lib/supervisor.js
@@ -118,6 +118,12 @@ function run(
     var stream = socket.pipe(split());
 
     stream.on("data", function(data) {
+      // In watch mode, the socket is drained which results in an extraneous
+      // message being sent. If we receive no data, ignore it.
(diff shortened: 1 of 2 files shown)


rtfeldman:
Published as `0.18.12`!

PR #245, opened 2018-01-17
DNS timeout should prevent downloading packages

michie1:
Yesterday I was running elm-test on the train without internet connection, which resulted in an error. 
The following pull request made it possible to work offline https://github.com/rtfeldman/node-test-runner/pull/232, but only takes the error code "ENOTFOUND" into account. With this pull request it will also take "EAI_AGAIN" into account, which means there is a DNS lookup timeout.


dlax:
Looks reasonable.
[dns.lookup()](https://nodejs.org/api/dns.html#dns_implementation_considerations) appears to use [getaddrinfo](https://linux.die.net/man/3/getaddrinfo) which documents `EAI_AGAIN` as "The name server returned a temporary failure indication. Try again later."

lib/elm-test.js
```diff
@@ -505,7 +505,7 @@ function generatePackageJson(filePathArgs) {
 function ensurePackagesInstalled(dir) {
   return new Promise(function(resolve, reject) {
     dns.lookup("elm-lang.org", function(err) {
-      if (err && err.code == "ENOTFOUND") {
+      if (err && (err.code == "ENOTFOUND" || error.code == "EAI_AGAIN")) {
```
dlax: Should be `err.code` (`error` is not defined).

michie1: Thanks! So embarrassing, I'm so happy there is a review process.

PR #246, opened 2018-01-25
fix typo file => filePath

eeue56:
in the error message below it's called `filePath`, which didn't exist


rtfeldman:
Thanks @eeue56!

PR #256, opened 2018-05-21
Fix encodeReasonType in JsonReporter

kazk:
Thanks for this project. We have some users [wanting to use Elm](https://github.com/Codewars/codewars-runner-cli/issues/520) and this will be very helpful.

This PR fixes the issue where failure reason's `type` is always set to `"custom"`, making it difficult to use the `data`.


rtfeldman:
Thanks @kazk!

PR #264, opened 2018-08-13
update fsevents and chokidar

m-shaka:
Hi. When using elm-test with node v10, [this error](https://github.com/strongloop/fsevents/issues/229) occurs. 

To fix this, I updated some dependencies.


rtfeldman:
@m-shaka Can you rebase on current `master` and pin the dependencies to exact versions? (no `^` in them, that is.)

Thanks for this!

halfzebra:
@m-shaka good stuff! 👍 

@rtfeldman what do you think of removing `package-lock.json` and setting `fsevents` and `chokidar` versions to "latest minor version"?

rtfeldman:
Thanks @m-shaka!

@halfzebra for applications I always like to pin dependencies exactly, so if two teammates install the same version they're always getting the same thing. Cuts down on "I can't reproduce; maybe uninstall and reinstall?"

Issue #242, opened 2017-12-30
Watch mode improvements


zwilias:
- [ ] **1**. #149 new/removed tests aren't picked up, and worse, throw an error
- [ ] **2**. #204 if initial compilation fails, `--watch` doesn't watch
- [ ] **3**. watch creates a new `new.Server` for every test run within a session
- [ ] **4**. filewatcher doesn't debounce and can trigger multiple testruns

I think a good way forward would be to make the supervisor have a single persistent `net.Server`, with the ability to `queue` testruns. This would be a first step towards taking care of point 4 and take care of point 3.

Untangling `generateAndRunTests` in `lib/elm-test.js` so that chokidar can do the filewatching independent of the compilation process would also be required. This would probably also involve making the `process.kill` at the end of the promise-chain in `lib/elm-test.js:runElmTest` check for `args.watch`. This would take care of points 1 and 2.

Finally, adding a `debounce` flag to the chokidar watcher would be the final piece to handling point 4; so that it's safer to switch branches with `elm-test --watch` running.

harrysarson:
I had a go at some of these issues. I did not touch the `net.Server` stuff but I think I might have fixed 1, 2 and 3.

harrysarson:
I believe these improvements have landed! Please do re-open (or make a new issue) if we can improve watch mode further!

lydell:
@harrysarson It sounds like you intended to close this issue but forgot?

PR #279, opened 2018-09-07
watch: prevent process exit, add debouncing, handle new/removed files.

### LINKED ISSUE #149: Renaming / removing test after --watch triggers an error
**Sequence to provoke the bug:**
- run `elm-test --watch`
- change the test file (eg. change a name of exposed test from `todoTest` to something else) and save
- `elm-test` will say:
```
Tests.elm changed. Rebuilding!
-- NAMING ERROR - /path/to/project/elm-stuff/generated-code/elm-community/elm-test/src/Mainf99326b294595b25aa7c60c9972d6ff6.elm

Cannot find variable `Tests.todoTest`.

13|     [     Test.describe "Tests" [Tests.todoTest] ]
                                     ^^^^^^^^^^^^^^
`Tests` does not expose `todoTest`. 

Detected errors in 1 module.                                        
Compilation failed for /path/to/project/elm-stuff/generated-code/elm-community/elm-test/src/Mainf99326b294595b25aa7c60c9972d6ff6.elm
```
ie. it thinks we still have the `todoTest`, when in reality we renamed it / deleted it / ...

This is because `generateAndRunTests` only gets run once, but it should get run in the callback of `watch` if the test files changed:
https://github.com/rtfeldman/node-test-runner/blob/master/bin/elm-test#L501-L509

Maybe something like:
```js
if (changedFileContainsTests) {
    generateAndRunTests(...);
} else {
    runTests(...);
}
```

### LINKED ISSUE #204: `--watch` flag does not work when an error occur on the first run
The test-runner should be listening for file changes even if the tests fail the first execution, right !?

harrysarson:
Fixes #149
Fixes #204

Relevant #242

This commit:
 - Makes some functions return promises so we know when they end.
 - Refactors the file watching logic to re-run more of the testing code.
 - Moves some logging calls around to keep output as it currently is.

**Note**: The tests fail on master so I don't know if they fail _more_ after this commit (I hope not :))

### DIFF
--- a/lib/elm-test.js
+++ b/lib/elm-test.js
@@ -2,6 +2,7 @@
 
 var packageInfo = require("../package.json");
 var pipeFilename = require("./pipe-filename.js");
+var chalk = require("chalk");
 var Version = require("./version.js");
 var Install = require("./install.js");
 var dns = require("dns");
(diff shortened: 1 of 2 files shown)


rtfeldman:
Thanks @harrysarson! I'm working on fixing the `master` tests, but I'll merge this in the meantime.

PR #299, opened 2018-10-09
Add console clearing on watch mode

### LINKED ISSUE #296: Clear screen in watch mode
A small feature request that I'd be very happy to PR if you'd like it - I find it useful how Jest clears the screen before each watch test run, so you don't have to always end up looking at the bottom of your terminal.

I think we could steal the clearing logic from Jest and insert it into the test runner. Happy to work on it if you'd like it :)

jackfranklin:
https://cl.ly/f0ad6cd54e24/sd;lfk.mov

I took this code directly from create-react-app and Jest, it's how they clear the screen.

I'm not sure what the white flashing is on the clear, but I think it's my config rather than this, but haven't dug too much. It's much more pronounced on the screencast than it actually is.

@rtfeldman do you think we want a flag to enable/disable this behaviour?

closes #296

### DIFF
--- a/lib/elm-test.js
+++ b/lib/elm-test.js
@@ -315,6 +315,12 @@ if (args._[0] === "make") {
   const generatedSrc = returnValues[1];
   const sourceDirs = returnValues[2];
 
+  function clearConsole() {
+    process.stdout.write(
+      process.platform === 'win32' ? '\x1B[2J\x1B[0f' : '\x1B[2J\x1B[3J\x1B[H'
+    );
+  }
+
   function run() {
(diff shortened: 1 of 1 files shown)


rtfeldman:
Looks good to me, thanks @jackfranklin! I'd like to leave it without a flag and see if anyone requests different behavior.

The CI failure is a known issue on `master`, so merging as-is!

rtfeldman:
I'll include this in the next beta release, once I've fixed the CI issue.

rtfeldman:
Published in `0.19.0-beta10!`

Issue #295, opened 2018-10-01
Elm-test hangs indefinitely 


smucode:
Perhaps every 20th build on our CI, elm-test (0.19.0-beta9) hangs. It just stops right after these messages: 

```
elm-test --verbose
Running elm make /builds/<snip path>/FooTest.elm <snip more tests>/builds/<snip path>/BarTest.elm --output /dev/null
Running elm make /builds/<snip path>/elm-stuff/generated-code/elm-explorations/test/src/Test/Generated/Main2122303807.elm --output /builds/<snip path>/elm-stuff/generated-code/elm-explorations/test/elmTestOutput.js
 ```

Aborting and retrying the tests works.

This was not a problem with 0.18 and the previous elm-test version.

We have no more than 100 test cases in 10 files.

**Comment by rtfeldman** [comment]
Thanks! I'm assuming this is a duplicate of https://github.com/rtfeldman/node-test-runner/issues/288 but I appreciate having a separate issue - more data is better! 👍

**Comment by rtfeldman** [comment]
I think c123c8 introduced this, but I've had no luck isolating why. It only seems to happen on Travis CI, not on local Linux systems. I'm gonna try rewriting that commit incrementally, pushing to Travis whenever I get it to a state where it works locally again, and see if I can isolate the part of that refactor that introduced the problem.

**Comment by bruckner** [comment]
I'm hitting this while running locally (not in CI) with elm-test 0.19.0-beta9 on macos 10.14 Mojave with npm 5.4.2.

Fresh install, ran a half dozen times no problem, now hanging every run.  Small test suite with two tests, was running sub-second.

**Comment by rtfeldman** [comment]
@brucker would it be possible for you to share the tests in question?

**Comment by bruckner** [comment]
@rtfeldman It was the tests copied below.  I downgraded to beta8 and that worked.  Just upgraded to beta10 and so far so good.
```
suite : Test
suite =
    describe "RelativeTime"
        [ describe "formatTime"
            [ test "Formats midnight correctly" <|
                \_ ->
                    let
                        midnight =
                            millisToPosix 0

                        result =
                            RelativeTime.formatTime utc midnight
                    in
                    Expect.equal "12:00 am" result
            , test "Formats noon correctly" <|
                \_ ->
                    let
                        noon =
                            millisToPosix (12 * 3600 * 1000)

                        result =
                            RelativeTime.formatTime utc noon
                    in
                    Expect.equal "12:00 pm" result
            ]
        ]
```

Issue #293, opened 2018-09-22
Let `--report=json` also have effect on `elm-test make`


ymtszw:
As of #284 we can "just compile" files under `tests/` with `elm-test make` which is quite nice!

However, currently `--report=json` option only has effect on reporting test results in JSON (tested on beta9), not for `elm make` compile results under the hood of `elm-test make`.

Since editor plugins are most likely depending on machine-parsable JSON format compile results, it would be essential to have `--report=json` also work for compile results.

Ideas could be:
- Just let currently existing `--report=json` to have effect both on compile results and test results. So that with `--report=json`, everything reported in JSON. For compilation results, just ignore `--report=junit`.
- Implicitly consider `--report=json` for test and compile separately. In `elm-test make`, `--report=json` has effect on compile (obviously). For non-`make`, `--report=json` affects on test results, but not compile results (current behavior)
- Explicitly separate options for test and compile. Lets say existing `--report` for test results, and new `--make-report` or `--compile-report` for compile results. This allows specifying different report mode for compile and tests. This is most explicit and non-ambiguous, and possibly useful if some consuming plugins are using pre-existing test result handling library that expects junit format.

Issue #288, opened 2018-09-15
0.19.0-beta9 freezes on Linux CI


Y0hy0h:
I didn't pin the `elm-test` version on my CI, and thus tested the fresh `0.19.0-beta9` update. Unfortunately, the tests [simply freeze after the statement on how to reproduce the tests](https://travis-ci.org/Y0hy0h/breakdown/builds/428983514#L686).

I reverted to `0.19.0-beta8` in [this PR](https://github.com/Y0hy0h/breakdown/pull/39) and it works again.

Before that, I tried:

- Resetting all caches (did not work)
- Executing `0.19.0-beta9` locally on Windows (worked!)

I'm completely unsure what the issue is and whether it might be something in my (unusual) setup. So please consume this issue with a pinch of salt (as Germans like to say).

rtfeldman:
Seems to be Linux-specific: https://travis-ci.org/rtfeldman/node-test-runner/builds/429400712

Y0hy0h:
I added to the description that I tested this locally on Windows and it worked.

duncanmalashock:
I ran into this also:
```
The command "elm-format --validate src --elm-version=0.19" exited with 0.
$ elm-test --verbose
Running elm make /home/travis/build/battermann/Tonal/tests/PlaceholderTests.elm --output /dev/null
Running elm make /home/travis/build/battermann/Tonal/elm-stuff/generated-code/elm-explorations/test/src/Test/Generated/Main2921126987.elm --output /home/travis/build/battermann/Tonal/elm-stuff/generated-code/elm-explorations/test/elmTestOutput.js
elm-test 0.19.0-beta9
---------------------
Running 1 test. To reproduce these results, run: elm-test --fuzz 100 --seed 100716193812984 /home/travis/build/battermann/Tonal/tests/PlaceholderTests.elm
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

magopian:
Can confirm that I had the same issue, and that downgrading to beta8 fixed it!

duncanmalashock:
> Can confirm that I had the same issue, and that downgrading to beta8 fixed it!

The downgrade worked for me too

rtfeldman:
[...]

PR #300, opened 2018-10-10
Limit the number of workers on CI

kazk:
@rtfeldman
This is the workaround from https://github.com/rtfeldman/node-test-runner/issues/288#issuecomment-427723129 for #288 and #295. From what I've found so far, macOS on Travis doesn't hang because it only has 2 cores instead of 48 of Linux.

Is this good enough for now to release next beta? I'd like to use #293 soon :)

As I wrote in the comment, one test is broken, but it should be a separate issue.
This build https://travis-ci.org/rtfeldman/node-test-runner/builds/429400712 from your comment 23 days ago shows macOS with Node.js 8 is passing, but I think [the test is actually failing](https://travis-ci.org/rtfeldman/node-test-runner/jobs/429400717#L298-L313) there as well.

image

### DIFF
--- a/lib/elm-test.js
+++ b/lib/elm-test.js
@@ -62,7 +62,8 @@ var args = minimist(process.argv.slice(2), {
   boolean: ["warn", "version", "help", "watch"],
   string: ["compiler", "seed", "report", "fuzz"]
 });
-var processes = Math.max(1, os.cpus().length);
+const isCI = process.env.CI === "true";
+var processes = Math.max(1, isCI ? 4 : os.cpus().length);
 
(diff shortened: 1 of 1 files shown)


kazk:
> I'll include this in the next beta release, once I've fixed the CI issue.

I probably misunderstood "the CI issue" in this comment.
Please close this if this isn't blocking. (I also just realized I should've done `Math.min(4, os.cpus().length)` instead.)

Sorry for the noise.

rtfeldman:
I found a proper fix, which is out in `beta10`. Thanks! ❤️

PR #302, opened 2018-10-11
Read files in parallel

kazk:
@rtfeldman I noticed `fs.writeFile`'s callback was missing when I looked at #301. This resolved the hanging on Travis. I can't test "Linux desktop with a Ryzen CPU", but this might be the cause.

https://github.com/rtfeldman/node-test-runner/blob/1bd4416883bfd943bf704d390422372f53323287/lib/Generate.js#L19-L30

### DIFF
--- a/lib/Generate.js
+++ b/lib/Generate.js
@@ -12,28 +12,44 @@ const path = require("path"),
   pipeFilename = require("./pipe-filename.js");
 
 function prepareCompiledJsFile(dest/*:string*/) {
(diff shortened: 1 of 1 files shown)


kazk:
Travis is only failing with Node 6 like `master` with this fix. https://travis-ci.org/kazk/node-test-runner/builds/440333018

rtfeldman:
Nice, thank you @kazk!

rtfeldman:
As an aside, I have no idea how the version without the callback was working in macOS, working on some Linux environments (e.g. Luke's Chromebook, including running the same version of Ubuntu that Travis ran) and failing on systems with more than 4 physical cores (e.g. Travis, my Ryzen machine).

I didn't use a callback because I mistakenly thought the `fs-extra` version of `readFile` returned a Promise, but surely whether it did or didn't wouldn't depend on number of cores...I still have no explanation for those symptoms.

PR #303, opened 2018-10-11
Fix --watch test

kazk:
- Use `byline` to ensure we get a `line`
- Use `strip-ansi` to make `runStart` event a valid JSON line and prevent terminal clearing when running tests locally
- Fix `hasTriggered` condition to wait for next line
- Fix to not send error so that Mocha doesn't get confused (both `done()` and `done(error)` was called). This was causing the weird test output showing this test as failure while having zero exit code.


kazk:
It's all green on Travis. https://travis-ci.org/kazk/node-test-runner/builds/440407581

PR #305, opened 2018-10-19
simplify consecutive pipe on List.map

teresy:
I noticed this simplification is possible (based on [elm-lint](https://package.elm-lang.org/packages/jfmengels/elm-lint/2.0.1/Lint-Rules-SimplifyPiping))


rtfeldman:
These are downloaded packages that we use for test fixtures, so I don't think we should edit them directly! 😅

PR #306, opened 2018-10-24
Only care about tests/ if no globs given

jackfranklin:
We run our tests via `elm-test 'frontend/**/*Tests.elm'` as our app is a
big monorepo and we don't want our Elm tests in the top level `tests`
directory.

However, doing this gives us an error that the `tests/` folder doesn't
exist. I updated the code to only check for and use `tests/` if not
given a glob.

@rtfeldman let me know what you think!

### DIFF
--- a/lib/Generate.js
+++ b/lib/Generate.js
@@ -59,14 +59,19 @@ function generateElmJson(projectRootDir/*:string*/, testRootDir/*:string*/, path
   const elmJsonPath = path.resolve(path.join(projectRootDir, "elm.json"));
   const generatedSrc = path.join(generatedCodeDir, "src");
 
-  if (!fs.existsSync(testRootDir)) {
-    console.error("Error: " + testRootDir + " does not exist. Please create a tests/ directory in your project root!");
-    process.exit(1);
-  }
+  const hasBeenGivenFileGlobs = filePaths.length > 0
 
-  if (!fs.lstatSync(testRootDir).isDirectory()) {
-    console.error("Error: " + testRootDir + " exists, but it is not a directory. Please create a tests/ directory in your project root!");
-    process.exit(1);
+  // if we were given file globs, we don't need to check the tests/ directory exists
+  if (hasBeenGivenFileGlobs === false) {
+    if (!fs.existsSync(testRootDir)) {
+      console.error("Error: " + testRootDir + " does not exist. Please create a tests/ directory in your project root!");
+      process.exit(1);
+    }
+
+    if (!fs.lstatSync(testRootDir).isDirectory()) {
+      console.error("Error: " + testRootDir + " exists, but it is not a directory. Please create a tests/ directory in your project root!");
+      process.exit(1);
+    }
   }
 
   fs.mkdirpSync(generatedCodeDir);
@@ -135,7 +140,7 @@ function generateElmJson(projectRootDir/*:string*/, testRootDir/*:string*/, path
     function(src) {
       return path.resolve(path.join(projectRootDir, src));
     }
-  ).concat(testRootDir);
+  ).concat(hasBeenGivenFileGlobs ? [] : [testRootDir]);
 
   testElmJson["source-directories"] = [
     // Include elm-stuff/generated-sources - since we'll be generating sources in there.
--- a/lib/Generate.js
+++ b/lib/Generate.js
@@ -52,17 +52,15 @@ function readUtf8(filepath) {
   });
 }
 
-function generateElmJson(projectRootDir/*:string*/, testRootDir/*:string*/, pathToElmBinary /*:string*/, filePaths /*:Array<string>*/) {
+function generateElmJson(projectRootDir/*:string*/, testRootDir/*:string*/, pathToElmBinary /*:string*/, filePaths /*:Array<string>*/, hasBeenGivenCustomGlobs /*int*/) {
   const generatedCodeDir = path.resolve(
     path.join(projectRootDir, "elm-stuff", "generated-code", "elm-explorations", "test")
   );
   const elmJsonPath = path.resolve(path.join(projectRootDir, "elm.json"));
   const generatedSrc = path.join(generatedCodeDir, "src");
 
-  const hasBeenGivenFileGlobs = filePaths.length > 0
-
-  // if we were given file globs, we don't need to check the tests/ directory exists
-  if (hasBeenGivenFileGlobs === false) {
+    // if we were given file globs, we don't need to check the tests/ directory exists
+  if (hasBeenGivenCustomGlobs === false) {
     if (!fs.existsSync(testRootDir)) {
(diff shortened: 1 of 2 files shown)


rtfeldman:
This is interesting!

One of my design goals with `elm-test` is: "You can open up any project and run the command `elm-test` and have it run all the project's tests."

That said, I've heard a few times that the hard requirement creates problems for people incrementally adopting Elm in existing applications.

Something we've talked about as a long-term goal is integrating `elm-test` into `elm publish` such that it can run all your tests, *and* your dependencies' tests, and tell you if what you're about to publish will break someone else's project. I definitely do not want to sacrifice that!

However, that concern only applies to packages, not applications. So here's how I'd like to proceed:

1. This is fine for applications. If `elm.json` has `"type": "application"` then it can use the glob as an alternative to `tests/`. It'll be less ergonomic in the sense that you can no longer run `elm-test` by itself, but it's presumably not the end of the world to invoke `elm-test` with the appropriate glob some other way.
2. I'd like packages to continue having the `tests/` requirement. I don't see a use case for incrementally adopting a package, and I think having that requirement will pay off in the long run.

Thoughts?

jackfranklin:
> This is fine for applications. If elm.json has "type": "application" then it can use the glob as an alternative to tests/

Agree that this is good for applications and less good for packages 👍 

> It'll be less ergonomic in the sense that you can no longer run elm-test by itself,

The goal of this PR is to not do this - the goal of this PR is that `elm-test` still works and assumes the `tests/` directory, but _if_ you supply a glob, it no longer cares about it. I think that's the behaviour my changes create.

> I'd like packages to continue having the tests/ requirement. I don't see a use case for incrementally adopting a package, and I think having that requirement will pay off in the long run.

Agree with this. For a package you have complete control of the repo and so this requirement makes sense.

I will update the PR - thanks for your thoughts!

jackfranklin:
> the goal of this PR is that elm-test still works and assumes the tests/ directory, but if you supply a glob, it no longer cares about it. I think that's the behaviour my changes create.

I am wrong! This changed the behaviour I think of running just `elm-test` in an application, so I need to fix that.

sporto:
I quite like the original idea, why the need for a `/tests` folder in packages? Collocating the tests with the code is quite nice and it would be great if it is supported for both applications and packages.

jackfranklin:
@rtfeldman I think I've now updated the code here to work as you explained it above. Let me know what you think!

lib/Generate.js
```diff
@@ -52,26 +52,13 @@ function readUtf8(filepath) {
   });
 }
 
-function generateElmJson(projectRootDir/*:string*/, testRootDir/*:string*/, pathToElmBinary /*:string*/, filePaths /*:Array<string>*/) {
+function generateElmJson(projectRootDir/*:string*/, testRootDir/*:string*/, pathToElmBinary /*:string*/, filePaths /*:Array<string>*/, hasBeenGivenCustomGlobs /*int*/) {
```
rtfeldman: I'm assuming `/*int*/` was meant to be `/*:boolean*/` here! 😄 

```suggestion
function generateElmJson(projectRootDir/*:string*/, testRootDir/*:string*/, pathToElmBinary /*:string*/, filePaths /*:Array<string>*/, hasBeenGivenCustomGlobs /*:boolean*/) {
```

lib/elm-test.js
```diff
@@ -302,15 +302,15 @@ if (args._[0] === "make") {
         ? 'No tests found for the file pattern "' +
           fileGlobs.toString() +
           '"\n\nMaybe try running elm-test with no arguments?'
-        : "No tests found in the tests/ directory.\n\nNOTE: Make sure you're running elm-test from your project's root directory, where its elm.json lives.\n\nTo generate some initial tests to get things going, run elm-test init";
+        : "No tests found in the tests/ directory.\n\nNOTE: Make sure you're running elm-test from your project's root directory, where its elm.json lives.\n\nTo generate some initial tests to get things going, run elm-test init.\n\nAlternatively, if you have tests in a different directory, try calling elm-test with a glob: elm-test 'frontend-app/**/*Tests.elm'.";
```
rtfeldman: Maybe say `"Alternatively, if your application has tests"` instead? This advice won't work if they are working on a package!

rtfeldman:
> why the need for a /tests folder in packages? Collocating the tests with the code is quite nice and it would be great if it is supported for both applications and packages.

It's very easy to add flexibility and very painful to take it away once people have come to rely on it. In the application case there are different design constraints, in that people may not have total control over their directory structures. In the package case, they do.

So I'm convinced that "I am going to have a really hard time using `elm-test` in my application if this is how the directory structure has to be" justifies adding this flexibility, especially since applications aren't shared the way packages are.

However, I don't feel confident I grasp the full implications of that choice for packages. The consequences of making that change are serious; if I add the flexibility and then end up having to take it away, I will have caused a lot of pain by saying yes right now. That's why I want to leave packages as-is.

I hear you that removing the `tests/` requirement sounds nice for package authors, and I'll keep it in mind for the future! This does not close the door to making that change later.

lib/Generate.js
jackfranklin: If only I had a compiler! Good spot !

lib/elm-test.js
jackfranklin: Nice idea - updated 👍

jackfranklin:
@rtfeldman nice comments, thank you! Updated the PR :)

lib/elm-test.js
```diff
@@ -302,15 +302,15 @@ if (args._[0] === "make") {
         ? 'No tests found for the file pattern "' +
           fileGlobs.toString() +
           '"\n\nMaybe try running elm-test with no arguments?'
-        : "No tests found in the tests/ directory.\n\nNOTE: Make sure you're running elm-test from your project's root directory, where its elm.json lives.\n\nTo generate some initial tests to get things going, run elm-test init";
+        : "No tests found in the tests/ directory.\n\nNOTE: Make sure you're running elm-test from your project's root directory, where its elm.json lives.\n\nTo generate some initial tests to get things going, run elm-test init.\n\nAlternatively, if your application has tests in a different directory, try calling elm-test with a glob: elm-test 'frontend-app/**/*Tests.elm'.";
```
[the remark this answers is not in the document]
rtfeldman: I don't think it does, but that seems like a good idea. 👍

jackfranklin:
@mgold nice idea - just to check, is https://github.com/rtfeldman/node-test-runner/pull/306/commits/58db81a11f5a53b279179c819b9f2f975812fc67 what you had in mind?

rtfeldman:
I'm assuming this is what you had in mind @mgold; I'm gonna merge so people can try `beta11` over the weekend...got the last release blocker bug merged too, so I think this might be the last beta release!

mgold:
> I'm assuming this is what you had in mind

Yup!

Issue #275, opened 2018-08-30
0.19 beta crashes on Windows 10


Y0hy0h:
I'm having the [same issue as was present for 0.18](https://github.com/rtfeldman/node-test-runner/issues/255) with the beta for 0.19.

When I hit save in my editor (happens in both Atom and VSCode, on two different machines), the watch will output the following log:
```
[successful test run omitted]

Watching for changes...

..\src\Code.elm changed. Rebuilding!
Success! Compiled 3 modules.
Unhandled exception while running the tests: { Error: EINVAL: invalid argument, unlink '\\.\pipe\elm_test-12400'
    at Object.fs.unlinkSync (fs.js:1061:18)
    at Function.rimrafSync [as sync] (C:\Users\Y0hy0h\AppData\Local\Yarn\Data\global\node_modules\rimraf\rimraf.js:306:17)
    at Object.removeSync (C:\Users\Y0hy0h\AppData\Local\Yarn\Data\global\node_modules\fs-extra\lib\remove\index.js:4:17)
    at Object.run (C:\Users\Y0hy0h\AppData\Local\Yarn\Data\global\node_modules\elm-test\lib\supervisor.js:250:8)
    at ChildProcess.<anonymous> (C:\Users\Y0hy0h\AppData\Local\Yarn\Data\global\node_modules\elm-test\lib\elm-test.js:204:18)
    at emitTwo (events.js:126:13)
    at ChildProcess.emit (events.js:214:7)
    at ChildProcess.cp.emit (C:\Users\Y0hy0h\AppData\Local\Yarn\Data\global\node_modules\cross-spawn\lib\enoent.js:40:29)
    at maybeClose (internal/child_process.js:925:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:209:5)
  errno: -4071,
  code: 'EINVAL',
  syscall: 'unlink',
  path: '\\\\.\\pipe\\elm_test-12400' }
```

6f53961077b6b29793d3b3a0198f96617bfe6a37 is supposed to fix the issue, but I have installed `elm-test@0.19.0-beta5` and it is still happening.

Y0hy0h:
[...]

PR #309, opened 2018-10-24
Fix windows EADDRINUSE

rtfeldman:
Attempt to fix https://github.com/rtfeldman/node-test-runner/issues/275#issuecomment-418756561


rtfeldman:
Huh! So everything worked without https://github.com/rtfeldman/node-test-runner/commit/963f0de986d1eaa57afd6576d8590b70deb8c76c in the mix? As in, the original bug appears to be gone?

Y0hy0h:
I haven't yet tried getting rid of 963f0de986d1eaa57afd6576d8590b70deb8c76c, so no, for now it simply freezes before the issue in #275 can manifest.

rtfeldman:
Ah ok - I just reverted it on the branch, so let's see how it does without it!

Y0hy0h:
Thanks for reverting it for me. I'm a bit tired, so I wouldn't have done it myself today, but this makes it trivial to test.

And it appears that it works! I'll poke it a bit more, but it has run well a few times now.

rtfeldman:
Appveyor now reports all passing tests and then hangs, leading to a failure.

I'm gonna diagnose that separately; seems ok now!

Y0hy0h:
@rtfeldman It _does_ terminate without `--watch`. It still ignores <kbd>Ctrl</kbd>+<kbd>C</kbd> with `--watch`.

However, that is workable already. So thanks a lot for making this work! :)

rtfeldman:
Thanks for your patience!

For future reference, would you mind opening a different issue for the `Ctrl-C` issue?

It won't be a release blocker, but I'd like to keep track of it!

Issue #310, opened 2018-10-26
Set up prettier + eslint


jackfranklin:
Hey @rtfeldman , would you have any objection to me setting up Prettier and/or ESLint with some reasonable rules (that this project would mostly pass)? I think it would make contributing much easier :)

rtfeldman:
Sure, go for it!

harrysarson:
I think this can be closed as #314 is merged!

harrysarson:
Actually eslint hasn't been added yet so we can leave open.

lydell:
ESLint was added in https://github.com/rtfeldman/node-test-runner/pull/425 🎉

PR #314, opened 2018-10-30
Add Prettier

jackfranklin:


jackfranklin:
@rtfeldman the first part of https://github.com/rtfeldman/node-test-runner/issues/310, linting to follow but this first part was super easy :)

jackfranklin:
@rtfeldman I'm not sure why there is a failure on Appveyor (it's saying it cannot find files to run Prettier on) but it works on Travis - might need your help with that!

package.json
```diff
@@ -8,7 +8,9 @@
   },
   "scripts": {
     "flow": "flow",
-    "test": "flow check && mocha tests"
+    "test": "flow check && npm run prettier:check && mocha tests",
+    "prettier:check": "prettier 'lib/**/*.js' --list-different",
```
rtfeldman: Looks like the single quotes are the problem: https://github.com/prettier/prettier/issues/4561#issuecomment-392817015

Try it without single quotes:
```suggestion
    "prettier:check": "prettier \"lib/**/*.js\" --list-different",
```

If that works, we should give `prettier:write` the same treatment!

rtfeldman:
Nice, that did the trick!

PR #315, opened 2018-11-05
Fix bug adding tests/ to source-directories

jackfranklin:
This bug meant that running `elm-test "/not/default/path/*.elm"` would
error if `tests/` directory didn't exist, because we'd add it to the
`source-directories` of the new `elm.json` file we create.

This fix now ensures we only add the `source-directories` if:

- we're a package, and therefore using `tests/` for our tests
- we're called without a glob, and therefore are defaulting to `tests/`
- the `tests/` folder exists on disk. This covers the case where I call
`elm-test "tests/Foo.elm"` to only run one file.

If none of the above is true, we don't default to adding the `tests`
directory.

This is a small bug with the implementation of #306

### DIFF
--- a/lib/Generate.js
+++ b/lib/Generate.js
@@ -81,7 +81,18 @@ function generateElmJson(
   // if we were given file globs, we don't need to check the tests/ directory exists
   // this is only for elm applications, which people may need to introduce slowly into their apps
   // for packages, we stick with the existing behaviour and assume tests are in tests/ so do the check always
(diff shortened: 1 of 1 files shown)


jackfranklin:
@rtfeldman this is a hopefully straight forward bug fix :)

rtfeldman:
Ahh nice catch!

rtfeldman:
Published as `beta12` - fingers crossed this is the last beta! 🤞

PR #317, opened 2018-11-06
Skip writing extra compile error message for JSON reporter

kazk:
Fix stderr output like:
```
{"type":"compile-errors"}Compilation failed while attempting to build
```

---

`package-lock.json` and `fixtures/elm-home/0.19.0/package/versions.dat` changed when I ran tests locally.


rtfeldman:
Hm, presumably we don't just want this for JSON, but also for TestUnit XML, yeah?

kazk:
Does elm-make have reporter for them?
I ran into this issue because I was trying to parse the compiler errors in JSON that's written to stderr. In other words, it shouldn't affect them if there's no need to parse stderr.

Another option is to write compiler error for JSON reporter to stdout and keep these messages. This way, the consumer doesn't need to worry about handling stderr in addition to stdout.
For example, piping doesn't work for compile errors currently:

```
$ (>&2 echo '{"compile-errors": []}') | jq .
```

```
$ (echo '{"compile-errors": []}') | jq . 
{
  "compile-errors": []
}
```

harrysarson:
I think the root issue has been resolved (?). @kazk if you wanna take this forward please re-open. Just doing some spring cleaning (is spring somewhere in the world right?) and closing this very old PR.

Issue #320, opened 2018-11-15
validating transitive dependencies - elm/file


doanythingfordethklok:
I get this error when attempting to run tests on a package that includes `elm/file` as a dependency. 

```
elm-test internal error: got an unexpected result from 'elm make' when validating transitive dependencies.  Please report this at https://github.com/rtfeldman/node-test-runner/issues
```

UPDATE: I was able to resolve the above error by updating peer dependencies of elm/file like this.

```
"dependencies": {
        "elm/core": "1.0.1 <= v < 2.0.0",
        "elm/file": "1.0.0 <= v < 2.0.0",
        "elm/json": "1.1.0 <= v < 2.0.0"
    },
```

Now this is the error:

`This test failed because it threw an exception: "ReferenceError: File is not defined"`

SidneyNemzer:
Regarding `ReferenceError: File is not defined`, see https://github.com/elm/core/issues/1002

doanythingfordethklok:
I upgraded elm/core and the tests are now working. Thank you for the quick fix.  :)

PR #321, opened 2018-11-15
Polyfill for File

andys8:
### Issue

Tests failing with
> This test failed because it threw an exception: "ReferenceError: File is not defined"

### Related
@SidneyNemzer referenced `elm/core` using `File`. https://github.com/elm/core/issues/1002
Slack: https://elmlang.slack.com/archives/C0CJ3SBBM/p1542311071629600

### Versions
Node v10.13.0
elm-test 0.19.0-rev2

Solves #320

### DIFF
--- a/templates/before.js
+++ b/templates/before.js
@@ -48,3 +48,7 @@ if (typeof FormData === "undefined") {
     this._data.push(Array.prototype.slice.call(arguments));
   };
 }
+
+if (typeof File === 'undefined') {                                                                                                                                                             
+  File = function() {};                                                                                                                                                                        
+}


evancz:
I think the fix I mention in the other thread is better, and it makes this code not need to be added.

andys8:
Thanks @evancz for fixing the issue in `elm/core`. I leave it to the maintainers, if it makes sense to add this pull request:

:heavy_plus_sign:  Matches the existing polyfills for `window`, `FormData`, `FileList` and others, and could benefit in the future if there's a regression

:heavy_minus_sign:  Not necessary for the fix, because https://github.com/elm/core/commit/ad161458f8d72d5ce89110fc767266afa0e155e5

andys8:
@rtfeldman added it https://github.com/rtfeldman/node-test-runner/commit/fa58d23cdea250f1e3c33e648f264e01ad66bad7

PR #329, opened 2019-01-13
Change elm-package.json -> elm.json in .flowconfig

harrysarson:
This must be hanging around from elm 18.

Although maybe `elm.json` does not need to be ignored?
It has not been ignored previously.


rtfeldman:
Thanks!

PR #330, opened 2019-01-13
prevent command injection in elm-test install

harrysarson:
`which` is used to get the path to the elm executable which allows
using `execFileSync` instead of `execSync` preventing any escape
characters included in the path to the elm compiler from being
evaluated by the shell.

I cannot think of anyway this could be exploited in its current state
but passing unsanitized user input to `execSync` is advised against.

See: https://nodejs.org/api/child_process.html#child_process_child_process_execsync_command_options

### DIFF
--- a/lib/elm-test.js
+++ b/lib/elm-test.js
@@ -8,6 +8,7 @@ var Compile = require('./Compile.js');
 var Generate = require('./Generate.js');
 var dns = require('dns');
 var processTitle = 'elm-test';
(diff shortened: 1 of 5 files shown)


harrysarson:
Tests run locally, appveyor is failing on `binwrap-install` which I believe is unrelated to this PR.

rtfeldman:
Great idea, thanks!

PR #332, opened 2019-01-18
Use elm-explorations/test 1.2.0

rtfeldman:


example-application/elm.json
```diff
@@ -13,11 +13,14 @@
...
+            "elm-explorations/test": "1.2.0"
         },
         "indirect": {
             "elm/random": "1.0.0",
-            "elm/time": "1.0.0"
+            "elm/time": "1.0.0",
+            "elm/html": "1.0.0",
+            "elm/virtual-dom": "1.0.2"
+        }
         }
```
harrysarson: One of these semi-colons should not be here

rtfeldman: Thanks!

PR #334, opened 2019-01-29
Remove unused imports

harrysarson:
Does what is says on the tin. Not a ground breaking change but makes the code base slightly tidier 🙂


harrysarson:
What is the status here?

jackfranklin:
@harrysarson merged! Thank you :)

PR #335, opened 2019-01-29
Remove console.log

harrysarson:
This was originally added for debugging. I intended to remove it before
making a pull request but evidently forgot!


harrysarson:
What is the status on this one too 👼

jackfranklin:
@rtfeldman I'm assuming this change isn't big enough to warrant a release just for this, but can I leave it to you to cut a new release when you see fit? :)

harrysarson:
This change only effects test files so users won't see it in a release. (By which I mean there should be no need to cut a release.)

Issue #263, opened 2018-07-12
Improve diffing of Union types


harrysarson:
Hi, I have come accross a difficulty when testingUnion types.

I have a test along the lines of:

```elm
test "union type diff" <|
    \() ->
        Err { description = "An error", context = "Explanation of what went so wrong" }
            |> Expect.equal (Ok "Success.")
```

This results in the following output:
![image](https://user-images.githubusercontent.com/16308754/42658789-451a32c6-861e-11e8-8aad-784276caf29b.png)

The diff is not very helpful, futhermore the attempts to highlight the differing characters make it hard to read the error message.

I was wondering if it would be possible to detect the case where different union types are being compared and format this case differently. 

Sorry if I have opened this issue in the wrong repo.
Cheers.

mgold:
I don't think the text being presented is that bad, but wow is the text highlighting bad. Reproduced.

rtfeldman:
Fix published in `0.19.0-rev5`

PR #336, opened 2019-02-10
Prevent many small highlighted blocks in diff.

harrysarson:
By comparing the number of edges of highlighted blocks to the number
of non-highlighted characters (highlighting is the result of
diffing),  messy diffs with lots of small highlighted blocks are
disabled.

Diffs will only be displayed if there are more plain characters than there
are edges highlighted blocks in both the actual and expected strings.

Fixes https://github.com/rtfeldman/node-test-runner/issues/263.

### DIFF
--- a/src/Test/Reporter/Console/Format.elm
+++ b/src/Test/Reporter/Console/Format.elm
@@ -90,18 +90,51 @@ highlightEqual expected actual =
 
     else
         let
+            isHighlighted =
+                Highlightable.resolve
+                    { fromHighlighted = always True
+                    , fromPlain = always False
+                    }
+
+            edgeCount highlightedString =
+                let
+                    highlights =
+                        List.map isHighlighted highlightedString
+                in
+                highlights
+                    |> List.map2 Tuple.pair (List.drop 1 highlights)
+                    |> List.filter (\(lhs, rhs) -> lhs /= rhs)
+                    |> List.length
+
             expectedChars =
                 String.toList expected
 
             actualChars =
                 String.toList actual
+
+            highlightedExpected =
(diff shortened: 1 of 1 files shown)


harrysarson:
I have not added a test as there seemed to be no where to put it 😢. Here is a test I used to verify this change locally:

```elm
module Example exposing (suite)

import Expect
import Test exposing (..)


suite : Test
suite =
    describe "union type diff"
        [ test "1" <|
            \() ->
                Err { description = "An error", context = "Explanation of what went so wrong" }
                    |> Expect.equal (Ok "Success.")
        , test "2" <|
            \() ->

                Err { context = "Success" }

                    |> Expect.equal (Ok "Success.")

        , test "3" <|
            \() ->
                Err { context = "OhK3 S-u5c6c4e2s2s  4" }
                    |> Expect.equal (Ok "Success.")
        ]
```

Before:

image

After:

image

jackfranklin:
@harrysarson this is awesome and a great change. I think we should figure out some place to put that test you were using as your test case, as I think it'll probably be a useful reference. I'm also not quite sure where in the repo it could go though.

harrysarson:
I don't think the test runner has any tests of its elm code at all.

harrysarson:
I have found a place to put elm unit tests but I will make a new PR for that as my idea will require some other changes. I would say to merge this as is an I will add the tests soon. :)

jackfranklin:
It looks great to me, however I'm not familiar with this part of the codebase, so I think I'd like to let @rtfeldman have a look before merging :)

rtfeldman:
I dig it! This has been bothering me since I added it. 😄 

Thanks @harrysarson!

PR #339, opened 2019-02-20
Separate elm src and add unit tests

harrysarson:
In order to add a test for https://github.com/rtfeldman/node-test-runner/pull/336, this commit moves all elm source files from `//src` to `//elm/src`.

A unit test covering #336 has also been added at `//elm/tests/`.

The `//elm/elm.json` file is used when generating the source and for
running unit tests.

There is still `//elm.json` but this is now used only by `//tests/ci.json`.


harrysarson:
CI will fail because this pull requests contains a test that will only pass once #336 is merged.

avh4:
> There is still `//elm.json` but this is now used only by `//tests/ci.json`.

What does `/tests/ci.js` use it for?  Should it be switched to use `/elm/elm.json` ?

harrysarson:
> What does `/tests/ci.js` use it for?

 `//tests/ci.js` runs `elm-test` from the root directory. `elm-test` looks for an `elm.json` file in its current directory and tests files in a `test` directory.

I think we should move `//elm.json` but then we would also have to modify `//tests/ci.js`. (While we are at it I think it would be good to move `//tests/*.elm` into a separate directory).

Should I add a commit to this PR?

harrysarson:
I am thinking of a way to remove `//elm.json` but I will put that in a separate pull request if that is okay with you folk.

Is there anything else that needs doing here?

rtfeldman:
Looks good! Thanks @harrysarson!

PR #341, opened 2019-02-24
Remove only in test suite

harrysarson:
I added this `.only` when implementing https://github.com/rtfeldman/node-test-runner/pull/330, sorry!

The full test suite hasn't been running for a while.
It is a very good thing that elm-test gives a non zero exit code when `only` is used to prevent issues like this.

The test suite now fails, will follow up with a commit to fix this.

### DIFF
--- a/tests/flags.js
+++ b/tests/flags.js
@@ -130,7 +130,7 @@ describe('flags', () => {
       assert.notEqual(runResult.code, 0);
     });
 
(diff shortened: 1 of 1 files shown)


harrysarson:
cheeky bump :)

mgold:
I don't know what happened to @rtfeldman but I do have a commit bit and this seems as innocuous at it gets.

PR #346, opened 2019-03-11
Allow test to run for longer

harrysarson:
This should stop appveyor ci from failing.

Refs: https://ci.appveyor.com/project/rtfeldman/node-test-runner/branch/master/job/tdt1wtlooqxp0cx5


harrysarson:
The root cause of these failures is something else - I think the spawned `elm-test --watch` process is never killed.

Issue #348, opened 2019-03-12
File watching broken on windows


harrysarson:
To reproduce: run `$node ./bin/elm-test --watch tests/OnePassing.elm` on master branch (using windows).

This bug was introduced by 2049ed5 where chokidar was bumped to version 2.1.2.

(This is why the azure pipelines CI has gone red.)

harrysarson:
I have found our bug. It is related to this breaking change in chokidar@2:

https://github.com/paulmillr/chokidar/blob/master/CHANGELOG.md#chokidar-200-dec-29-2017

> Breaking: Upgrade globbing dependencies which require globs to be more strict and always use POSIX-style slashes because Windows-style slashes are used as escape sequences

PR #349, opened 2019-03-12
Fix broken file watch on windows

harrysarson:
Fix regression added in 2049ed5 where chokidar@2 silently
fails to watch globs when windows path seperators are present.

Fixes: https://github.com/rtfeldman/node-test-runner/issues/348
Refs: 2049ed5
Refs: https://github.com/paulmillr/chokidar/blob/master/CHANGELOG.md#chokidar-200-dec-29-2017

### DIFF
--- a/lib/elm-test.js
+++ b/lib/elm-test.js
@@ -258,21 +258,16 @@ function resolveGlobs(fileGlobs) {
   return _.flatMap(globs, resolveFilePath);
 }
 
-function getWatchPaths(projectRootDir /*:string*/) {
-  var watchedSourcePaths;
-  var elmJson = fs.readJsonSync(path.join(projectRootDir, 'elm.json'), 'utf8');
+function getWatchPaths(elmJson) {
+  let watchedSourcePaths;
   if (elmJson['type'] === 'package') {
-    watchedSourcePaths = ['./src'];
+    watchedSourcePaths = ['src'];
   } else {
     watchedSourcePaths = elmJson['source-directories'];
   }
-  var watchedTestPaths = path.join(projectRootDir, 'tests');
-  var watchedPaths = watchedSourcePaths
-    .concat(watchedTestPaths)
-    .map(function(sourcePath) {
-      return path.resolve(projectRootDir, sourcePath) + '/**/*.elm';
-    });
-  return watchedPaths;
+  return [...watchedSourcePaths, 'tests'].map(function(sourcePath) {
+    return path.posix.join(sourcePath, '**', '*.elm');
+  });
 }
 
 var report;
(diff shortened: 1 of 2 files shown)


harrysarson:
May I make a cheeky bump of this pull?

I believe that the latest published version of elm-test on npm has broken file watch on windows.

harrysarson:
Given it has been a week, I hope it is socially acceptable to bump this pull request again.

Have a lovely Friday!

harrysarson:
Another week has gone by so I guess I can bump this again 😄 

Please let me know if there is anything I can do to help this get merged.

mgold:
@rtfeldman ^^

lib/elm-test.js
```diff
@@ -258,21 +258,16 @@ function resolveGlobs(fileGlobs) {
...
   }
-  var watchedTestPaths = path.join(projectRootDir, 'tests');
-  var watchedPaths = watchedSourcePaths
-    .concat(watchedTestPaths)
-    .map(function(sourcePath) {
-      return path.resolve(projectRootDir, sourcePath) + '/**/*.elm';
-    });
-  return watchedPaths;
+  return [...watchedSourcePaths, 'tests'].map(function(sourcePath) {
+    return path.posix.join(sourcePath, '**', '*.elm');
```
rtfeldman: What happens if I run this from a directory that isn't the project root?

Before, it would join with the project root to make that not matter. Now it appears to work relative to the current directory, wherever that might be.

This seems like a potential regression!

harrysarson: Thanks for the review! In [line 452](https://github.com/rtfeldman/node-test-runner/pull/349/files/ba14ad1b33696edb8907a1a4e5f2d0c4b8e0c3f6#diff-047f238be7300455dd55a8c22225e53fR452) I set chokidar's working directory to `projectRootDir`.

Therefore, chokidar works relative to _its_ current directory which is the _project's_ root directory. Therefore, I believe this PR does not introduce a regression.

Saying that, it is clear that the intention of the code is not clear - I believe I can rename a variable or two to clarify which paths are relative to what.

harrysarson: @rtfeldman I went ahead and checked if running elm-test from a subdirectory of a project with my changes and found #355.

> What happens if I run this from a directory that isn't the project root?

Currently elm-test crashes due to a different bug!

harrysarson:
CI is green again :)

harrysarson:
Having been another week, I thought I might bump this again. Please tell me if there is something more constructive I can do than generating weekly notifications.

rtfeldman:
Cool, thanks for the fix and for your patience, Harry!

harrysarson:
Yay 🎉

PR #350, opened 2019-03-13
Update flow and prettier to latest versions

harrysarson:
Later versions of flow require the return type of some functions to be specified.

Only the most recent versions of prettier can process the extra type comments that are required. I also had to add the `--parser flow` command line argument to prettier to stop it messing around with the comment.

This pull request is the result of updating the two packages and making the minimal changes required for ci to pass.


harrysarson:
CI failure is unrelated - although it would be good to re run CI once #348 lands just to make sure.

harrysarson:
What is the status on these pull requests? They don't contain any urgent changes but they have been open for over two months now.

As always, if there is anything helpful I can do, I would be very happy to do so!

mgold:
@rtfeldman It might be time to give Harry a commit bit and permission to use it.

harrysarson:
Cheeky ping?

avh4:
Looks great!  I included this in https://github.com/rtfeldman/node-test-runner/pull/372

harrysarson:
Thanks!

PR #351, opened 2019-03-17
Tidy test fixtures

harrysarson:
### Background

In https://github.com/rtfeldman/node-test-runner/pull/339#issuecomment-465881613, @avh4 asked (paraphrased):

> What does` /tests/ci.js` use `/elm.json` for? Should it be switched to use /elm/elm.json?

The answer is that (until #339) landed, the elm source code and elm tests used the same `elm.json`. Now the source and tests use different `elm.json`'s but the `elm.json` used by tests is still hanging about in the root directory.

Additionally, test fixtures are also been stored in the root directory.

## This pull request

* Moves all the fixtures directory from the root into tests.
* Widens the definition of "test fixtures" to include elm source files, tests files and json files used to run our tests.
* Moves all these files (which are currently either in `/tests/` or in the project root) into `/tests/fixtures/`. 

### Notes

* This moves the `/elm.json` file into `/tests/fixtures/` as requested in #339.
* I have not touched `/test/init-test/` this directory does not seem to be used by any tests and has an `elm-package.json` file. It should be either be deleted or a test added that uses it. For now I am leaving it well alone.
* There is a lot of noise in the diff for this PR due to moving the generated files. I tried to separate the noises changes and changes of note so reviewing each commit separately may be the way forward.


mgold:
This commits a lot of files under `tests/fixtures/elm-home`. Will these be checked out by `elm install`? It would be nice to git ignore this directory.

harrysarson:
> This commits a lot of files under `tests/fixtures/elm-home`

It only commits files that were previously there under `/fixtures/elm-home` (these commits "move" rather than "add" files).

> Will these be checked out by `elm install`? 

This is the test runner package so cannot be `elm install`ed. As for `npm install`ing, the tests are listed in the `.npmignore` file and, as such, will *not* be installed. (A list of the files that are installed by npm: https://unpkg.com/elm-test@latest/). 

> It would be nice to git ignore this directory.

I agree. But I think that is a separate change from the one I am hoping to make here.

mgold:
Fair enough. Seems fine to me.

harrysarson:
Before this I merged and after #349 lands on master I will rebase to get green CI

harrysarson:
I have fixed the merge conflicts!

harrysarson:
This has 👍 from @mgold and @rtfeldman. Can we merge it?

Issue #273, opened 2018-08-27
Testing with incompatible packages results in unhelpful error


zgohr:
```
elm-test internal error: got an unexpected result from 'elm make' when validating transitive dependencies.  Please report this at https://github.com/rtfeldman/node-test-runner/issues
```

Changing to `elm-stuff/generated-code/elm-explorations/test` and running `elm make` gave the following result

```
-- INVALID PACKAGE DEPENDENCIES --------------------------------------- elm.json

The dependencies in your elm.json are not compatible.

Did you change them by hand? Try to change it back! It is much better to add
dependencies with elm install or the dependency management tool in elm reactor.

Please ask for help on the Elm slack <http://elmlang.herokuapp.com/> if you try
those paths and still cannot figure it out!
```

Resolving the dependency mismatch resolved the error, but passing through this error message would have reduced much confusion.

rtfeldman:
@zgohr Can you paste the `elm.json` that led to this error?

zgohr:
```
{
    "type": "package",
    "name": "cuducos/elm-format-number",
    "summary": "Format numbers as pretty strings",
    "license": "BSD-3-Clause",
    "version": "6.0.0",
    "exposed-modules": [
        "FormatNumber",
        "FormatNumber.Locales"
    ],
    "elm-version": "0.19.0 <= v < 0.20.0",
    "dependencies": {
        "elm/core": "1.0.0 <= v < 2.0.0",
        "myrho/elm-round": "1.0.2 <= v < 2.0.0"
    },
    "test-dependencies": {
        "elm-explorations/test": "1.0.0 <= v < 2.0.0"
    }
}
```

andys8:
Another example:

```json
{
    "type": "application",
    "source-directories": [
        "src"
    ],
    "elm-version": "0.19.0",
    "dependencies": {
        "direct": {
            "NoRedInk/elm-json-decode-pipeline": "1.0.0",
            "Punie/elm-reader": "1.0.0",
            "arturopala/elm-monocle": "2.1.0",
            "ccapndave/elm-update-extra": "4.0.0",
            "cuducos/elm-format-number": "6.0.2",
            "danyx23/elm-uuid": "2.1.2",
            "elm/browser": "1.0.1",
            "elm/core": "1.0.2",
            "elm/http": "1.0.0",
            "elm/json": "1.1.2",
            "elm/time": "1.0.0",
            "elm/url": "1.0.0",
[...]

PR #352, opened 2019-03-27
Fix elm test when source package has no dependancy on elm/core

harrysarson:
These commits:

* Update the test runner's elm dependencies to fix @Maxim-Filimonov SSCCE.
* Add the sscce as a test.

---

Note that this does not fix the underlying issue of invalid dependencies giving an unhelpful error.

Refs: #273

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -44,6 +44,7 @@ example-application/
 example-application-no-tests/
 example-package/
 example-package-no-tests/
+example-package-no-core/
 
 README.md
 CHANGELOG.md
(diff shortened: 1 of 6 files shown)


harrysarson:
Linux CI: https://travis-ci.com/harrysarson/node-test-runner/builds/105985224

rtfeldman:
I've been trying to think if this could cause problems with applications that explicitly depend on `elm/core` 1.0.0 and `elm/json` 1.0.0, but I'm coming up blank.

So...I *think* this is okay!

Thanks @harrysarson!

PR #356, opened 2019-04-27
Leverage `elm-json solve`

### LINKED ISSUE #273: Testing with incompatible packages results in unhelpful error

### LINKED ISSUE #277: elm-test can be tricked into generating invalid dependencies
If `elm.json` has too-wide dependency ranges, it is possible that building normally with `elm make` will work but running the tests with `elm-test` will not. For example, consider the following `elm.json`:

```json
{
    "type": "package",
    "name": "ianmackenzie/elm-test-dependencies",
    "summary": "Tricking elm-test into specifying invalid dependencies",
    "license": "MPL-2.0",
    "version": "1.0.0",
    "exposed-modules": [
        "Module"
    ],
    "elm-version": "0.19.0 <= v < 0.20.0",
    "dependencies": {
        "elm/core": "1.0.0 <= v < 2.0.0",
        "ianmackenzie/elm-float-extra": "1.0.0 <= v < 2.0.0"
    },
    "test-dependencies": {
        "elm-explorations/test": "1.0.0 <= v < 2.0.0"
    }
}
```

`ianmackenzie/elm-float-extra` currently only has two versions - 1.0.0 for Elm 0.18 and 1.0.1 for Elm 0.19. `elm make` will end up installing 1.0.1, but the `elm.json` generated by `elm-test` will use a version constraint of 1.0.0 instead. Since that version only exists for Elm 0.18, `elm make` will produce a "The dependencies in your elm.json are not compatible" message, which ends up getting reported by `elm-test` as "got an unexpected result from 'elm make' when validating transitive dependencies".

The fix was to use the correct version range `"ianmackenzie/elm-float-extra": "1.0.1 <= v < 2.0.0"` in the package `elm.json`. This shouldn't happen if `elm install` is used to install dependencies, since it will only set up valid ranges. I only ran into it since I just copied my old `elm-package.json` to `elm.json` and then modified it as necessary. But it seems like a fairly easy mistake to make if for whatever reason you're editing `elm.json` files by hand - I know that on Slack at least @sporto ran into the same issue.

If it helps, I've made [a skeleton project](https://github.com/ianmackenzie/elm-test-dependencies) that reproduces the issue - cloning and running `elm-test` should trigger the "unexpected result" error message, and fixing the dependency constraint as described above should get everything working.

I'm on Windows 10 using `elm-test@0.19.0-beta6`.

### LINKED ISSUE #323: The dependencies in your elm.json are not compatible.
Just installed the elm-test cli via npm and ran `elm-test init`, and am now getting the following error.

`The dependencies in your elm.json are not compatible.`

I have not edited my elm.json by hand and my actual code compiles fine using `elm make`.

elm-test version 0.19.0-rev3

My elm.json:

```
{
    "type": "application",
    "source-directories": [
        "src"
    ],
    "elm-version": "0.19.0",
    "dependencies": {
        "direct": {
            "elm/browser": "1.0.1",
            "elm/core": "1.0.2",
            "elm/file": "1.0.1",
            "elm/html": "1.0.0",
            "ericgj/elm-csv-decode": "2.0.1",
            "lovasoa/elm-csv": "1.1.5",
            "mdgriffith/elm-ui": "1.1.0"
        },
        "indirect": {
            "elm/bytes": "1.0.3",
            "elm/json": "1.1.1",
            "elm/regex": "1.0.0",
            "elm/time": "1.0.0",
            "elm/url": "1.0.0",
            "elm/virtual-dom": "1.0.2"
        }
    },
    "test-dependencies": {
        "direct": {
            "elm-explorations/test": "1.2.0"
        },
        "indirect": {
            "elm/random": "1.0.0"
        }
    }
}
```

### LINKED ISSUE #328: Internal error result from 'elm make'
Using: `0.19.0-rev3`

elm.json
```json
{
    "type": "package",
    "name": "romariolopezc/hmac-sha1",
    "summary": "Compute HMAC with SHA-1 hash function",
    "license": "MIT",
    "version": "1.0.0",
    "exposed-modules": [
        "HmacSha1"
    ],
    "elm-version": "0.19.0 <= v < 0.20.0",
    "dependencies": {
        "TSFoster/elm-sha1": "1.0.0 <= v < 2.0.0",
        "elm/bytes": "1.0.7 <= v < 2.0.0",
        "elm/core": "1.0.0 <= v < 2.0.0",
        "ktonon/elm-word": "2.1.2 <= v < 3.0.0",
        "waratuman/elm-coder": "3.0.0 <= v < 4.0.0"
    },
    "test-dependencies": {
        "elm-explorations/test": "1.2.0 <= v < 2.0.0"
    }
}
```

I did:
```bash
$ yarn global add elm-test
$ elm-test init
$ elm-test
elm-test internal error: got an unexpected result from 'elm make' when validating transitive dependencies.  Please report this at https://github.com/rtfeldman/node-test-runner/issues
```

### LINKED ISSUE #374: Indirect elm/json overwritten with direct elm/json with different bad version.
The indirect elm/json version has been changed from 1.1.3 to 1.0.0 as a direct dependency when the elm-test elm.json file is generated, which results in an “INVALID PACKAGE DEPENDENCIES” error.

Elm discourse link: https://discourse.elm-lang.org/t/elm-test-how-to-fix-package-dependency-errors/4054

This elm.json:

```
{
    "type": "application",
    "source-directories": [
        "src"
    ],
    "elm-version": "0.19.0",
    "dependencies": {
        "direct": {
            "elm/browser": "1.0.1",
            "elm/core": "1.0.2",
            "elm/html": "1.0.0",
            "elm/http": "2.0.0",
            "elm/parser": "1.1.0"
        },
        "indirect": {
            "elm/bytes": "1.0.8",
            "elm/file": "1.0.5",
            "elm/json": "1.1.3",
            "elm/time": "1.0.0",
            "elm/url": "1.0.0",
            "elm/virtual-dom": "1.0.2"
        }
    },
    "test-dependencies": {
        "direct": {
            "elm-explorations/test": "1.2.2"
        },
        "indirect": {
            "elm/random": "1.0.0"
        }
    }
}
```
Is transformed into the generated elm.json:

```
{
    "type": "application",
    "source-directories": [
        "/home/rupert/tmp/elm-test/elm-stuff/generated-code/elm-explorations/test/src",
        "/home/rupert/.npm-global/lib/node_modules/elm-test/elm/src",
        "/home/rupert/tmp/elm-test/src",
        "/home/rupert/tmp/elm-test/tests"
    ],
    "elm-version": "0.19.0",
    "dependencies": {
        "direct": {
            "elm/core": "1.0.2",
            "elm/json": "1.0.0",
            "elm/random": "1.0.0",
            "elm/time": "1.0.0",
            "elm/browser": "1.0.1",
            "elm/html": "1.0.0",
            "elm/http": "2.0.0",
            "elm/parser": "1.1.0",
            "elm-explorations/test": "1.2.2"
        },
        "indirect": {
            "elm/bytes": "1.0.8",
            "elm/file": "1.0.5",
            "elm/url": "1.0.0",
            "elm/virtual-dom": "1.0.2"
        }
    },
    "test-dependencies": {
        "direct": {},
        "indirect": {}
    }
}
```

zwilias:
Fixes #328, fixes #277, fixes #273, fixes #323, fixes #374

I recently released [`elm-json`](https://github.com/zwilias/elm-json), a tool for interacting with `elm.json` files. One of the available subcommands is `solve`: Given an elm.json file (for a package or application), it tries to generate a single set of direct and indirect dependencies that satisfy all constraints. It has 2 flags that make it useful for `node-test-runner`:
- `--test` => Includes the `test-dependencies` and promotes them to direct dependencies
- `--extra` => Allows providing an extra set of packages that need to be direct dependencies of the final result

Since `elm-json` can be installed through `npm` and has binaries available for OS X, Windows and Linux (statically linked), it seems like a fine candidate to use in `node-test-runner` for calculating the final set of dependencies needed for the generated application.

---

I've tested this on all the failure cases listed in the linked tickets, as well as a number of my own packages and applications. As far as I can tell, it _just works_.

With this in place, we could also start thinking about running tests for packages for both the minimal allowed versions (by passing the `--minimize` flag to `elm-json solve`) and the maximal allowed versions. I did not include such exploration in this PR, but want to note that this now becomes feasible.

---

Cons of this approach:
- extra (binary) dependency (_but we already have `elmi-to-json`, so that's not a blocking factor_)
- very young package (_though I suspect the `elm-json solve` interface will not change in the foreseeable future_)
- depending on a tool written in Rust increases the barrier for dealing with issues related to interactions with `elm-json`


avh4:
> Cons of this approach

I think another notable one is that `elm-json` is written in rust -- I think haskell and javascript are are already reasonably involved in the elm ecosystem, but having a very different language like rust required for a central tool like elm-test add notable risk.  (Specifically, if there are issues in elm-test due interactions with elm-json, there is a very large barrier of entry for someone to be able to debug both projects.)

> With this in place, we could also start thinking about running tests for packages for both the minimal allowed versions and the maximal allowed versions

That sounds cool!

avh4:
(Also, I'm not sure why the Travis-CI build isn't running for this PR)

zwilias:
> having a very different language like rust required for a central tool like elm-test add notable risk

This is true! To be fair, there are [experiments](https://github.com/rtfeldman/node-test-runner/tree/rust-cli) with actually rewriting the elm-test CLI in Rust, so I feel fairly confident that a reasonable amount of people who could potentially debug issues should they arise, are involved with the `node-test-runner` project. On the other hand, that experiment might have been abandoned for exactly this reason 😅 

> (Also, I'm not sure why the Travis-CI build isn't running for this PR)

I have no idea why, but travis was disabled for PR's. I've re-enabled it after I noticed it wasn't running. I still have a commit pending for removing the `version.js` file, so 🤞 that the tests will trigger when I push that!

harrysarson:
> makes #352 redundant

I do believe that #352 is orthogonal to this and could both be merged. #352 updates the packages that the test runner uses and adds some tests.

zwilias:
@harrysarson The tests make sense, it's the changes to that `elm.json` file that are made redundant. I believe that file isn't used any longer as of this PR. In that sense, this PR fixes the same issue - it ensures a compatible version of `elm/core` is installed, even when missing in the package under test.

Perhaps I should remove that `elm.json` file from the repo in this PR as well. I'll have a look at doing so 👍

harrysarson:
> Perhaps I should remove that elm.json file from the repo in this PR as well.

If that is possible it would be a very good idea to do so. Keeping this file up to date is a pain.

zwilias:
> If that is possible it would be a very good idea to do so.

I'm less convinced. As a means of "propagating constraints" - yeah, that's painful and shouldn't be happening. But as an elm.json that's useful when actually working on the Elm code, it seems useful to have. It shouldn't be harder to keep up to date than any other elm.json.

@rtfeldman thoughts?

harrysarson:
To unblock this, I propose:

* `elm-test` detects if `elm-json` is installed.
* If so, it uses `elm-json` as per this PR.
* Otherwise, it uses the existing logic (the logic that this pull request deletes).
* If the existing logic fails, we append a little hint to the error message suggesting that the user can install `elm-json`.

Thoughts?

harrysarson:
Bump :)

harrysarson:
Bump :)

mpizenberg:
One important cons that I see is that `elm-json solve` needs access to the web and takes a lot of time (I guess because I have poor connectivity here). Is it possible to have an offline version that tries to solve constraints with already installed packages?

harrysarson:
That is a good point.

Saying that, elm-test runs the elm compiler under the hood so running it offline is not guaranteed to work (i.e. the elm compiler may decide it needs to download packages which requires internet access).

harrysarson:
@zwilias could you chime in on this: 

> One important cons is that elm-json solve needs network access and takes a lot of time (I guess because I have poor connectivity here). This would prevent usage of elm-test in an offline context. Is it possible to have an offline version that tries to solve constraints with already installed packages?

I think you recently released a new version of elm-json that solves this issue.

---

Regarding this pull request, I now have commit rights to this repository! The cons:

> Cons of this approach:
> 
> -   extra (binary) dependency (_but we already have `elmi-to-json`, so that's not a blocking factor_)
> -   very young package (_though I suspect the `elm-json solve` interface will not change in the foreseeable future_)
> -   depending on a tool written in Rust increases the barrier for dealing with issues related to interactions with `elm-json`

are, I think, worth accepting for the improvements this PR brings. Especially because elm-json is no longer so young and (subjectively) I think rust knowledge has increases across the elm community.

Therefore, I am inclined to **merge** this (after a rebase) in **two to four weeks time** unless anyone raises an further objections before then!

zwilias:
There is indeed an offline mode, however, `elm-json` also just does a whole lot less network io now (by caching `elm.json` files where possible). Even on bad connections, using elm-json should be very much feasible.

It might potentially be interesting to look into caching the generated `elm.json` and only regenerating if the input elm.json changed, but that feels like an optimization in elm-test land, rather than in `elm-json` :)

harrysarson:
Thanks @zwilias, that sounds great 👍 as far as I can tell there are no blockers to merging soon then.

lydell:
Since #451 was merged, we can close this one.

zwilias:
#451 did this, but better. Thanks @lydell!

PR #361, opened 2019-05-15
Tar vulnerability: run npm audit fix

mgold:
Fixes #360 if I've done it right.

Issue #360, opened 2019-05-14
High severity vulnerability


smucode:
`npm-audit` reveals `elm-test` has some vulnerable dependencies. Would be nice to get them upgraded so we can get rid of this error:

```
$ npm init -y && npm install --save elm-test && npm audit

...

+ elm-test@0.19.0-rev6
added 378 packages from 208 contributors and audited 2388 packages in 10.462s
found 1 high severity vulnerability
  run `npm audit fix` to fix them, or `npm audit` for details
                                                                                
                       === npm audit security report ===                        
                                                                                
# Run  npm install elm-test@0.19.0  to resolve 1 vulnerability
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ High          │ Arbitrary File Overwrite                                     │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ tar                                                          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ elm-test                                                     │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ elm-test > fsevents > node-pre-gyp > tar                     │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://npmjs.com/advisories/803                             │
└───────────────┴──────────────────────────────────────────────────────────────┘


found 1 high severity vulnerability in 2388 scanned packages
  run `npm audit fix` to fix 1 of them.

```

avh4:
The reported path is `elm-test > fsevents > node-pre-gyp > tar`, so it looks like this can be solved by updating `fsevents` to `^1.2.8` (upstream issue: https://github.com/fsevents/fsevents/issues/262)

smucode:
@mgold Possible to get this released? :)

mgold:
@rtfeldman ?

harrysarson:
I don't think all the vulnerabilities have been fixed.

rtfeldman:
Correct - we tried a fix for the `elm` package but the new approach runs into issues with the Windows virus scanner apparently, and I don't want to publish a new version that depends on it until that's resolved.

`elmi-to-json` does not have a fix yet.

harrysarson:
Probably worth reopening this issue then

stoeffel:
fyi. elmi-to-json is fixed.

harrysarson:
See #366!

jwoudenberg:
[...]

PR #363, opened 2019-05-31
Drop fsevents

harrysarson:
I am not 100% that we do not need fsevents - will see if CI passes.

Fixes: #361 
Fixes: #360


harrysarson:
@mgold what do you reckon?

mgold:
We never call `require('fsevents')` so this seems okay.

PR #366, opened 2019-06-17
bump elmi-to-json to latest

harrysarson:
Addresses (but does not fix) #360


rtfeldman:
Lovely, thanks @harrysarson!

PR #367, opened 2019-06-18
Clarify configuration section by renaming header

harrysarson:
This section describes the command line arguments to elm-test rather than configuration.
Some of the arguments (say `--seed`)  _do_ configure the test runner but others (such as help) do not.


mgold:
I'm going to say extensive review is not required for this one.

PR #370, opened 2019-07-12
upgrade lodash

jinjor:
A security vulnerability is reported by GitHub.

> Vulnerable versions: < 4.17.13
> Patched version: 4.17.13
>
> Affected versions of lodash are vulnerable to Prototype Pollution.
> The function defaultsDeep could be tricked into adding or modifying properties of Object.prototype using a constructor payload.


avh4:
Thanks!  I upgraded this and all the other dependencies in https://github.com/rtfeldman/node-test-runner/pull/372

PR #373, opened 2019-07-21
Add node 12 to CI builds

avh4:


appveyor.yml
```diff
@@ -1,11 +1,11 @@
 environment:
   ELM_VERSION: "0.19.0"
   matrix:
+  - nodejs_version: "12"
   - nodejs_version: "10"
   - nodejs_version: "8"
 
 platform:
-  # - x86
```
harrysarson: Out of interest: why do we not run tests on x86?

avh4: I'm assuming for the same reason I ran into in https://github.com/rtfeldman/node-elm-compiler/pull/98 -- that  elm@0.19.0-no-deps doesn't provide 32-bit Windows binaries, so I guess Evan doesn't want to support that platform any more?

harrysarson: Makes sense :+1:

harrysarson:
@avh4 we can merge I think!

Issue #359, opened 2019-05-09
--help exits non-zero


allenap:
I think it's usual that a program would exit zero (success) after being invoked with `--help`. I tried it out with `elm`, `ls`, `find`, and a few others – they all exit zero.

I use `elm-test --help` as a "liveness" test – and also to give it an opportunity to download binaries – so exiting zero would be useful to make my process more robust.

I think the following change will do it:

```diff
--- lib/elm-test.js
+++ lib/elm-test.js
@@ -125,7 +125,7 @@ if (args.help) {
     '[--watch] # Run tests on file changes',
   ].forEach(printUsage);

-  process.exit(1);
+  process.exit(0);
 }

 if (args.version) {
```

harrysarson:
@rtfeldman it has occurred to me that this is probably a semver major change. I have no idea how elm-test should follow semver though?

harrysarson:
I guess this change will be released with the 0.19.1 version of elm test which can reasonably containing breaking changes as `elm-test` is in 0.x versioning.

PR #377, opened 2019-08-03
fix tests that check exit code of spawned elm test

harrysarson:
Currently, a few tests check the `code` property of spawned processes.
However, the cross-spawn library does not define a `code` property on
the objects it returns; instead it defines a `status` property!

The test runner's unit tests where checking that the `code` was not
zero. Of course the code was not zero: it was `undefined` :angry:

As well as fixing the unit tests to check `.status` instead of `.code`
this commit fixes a unit test checked (or more accurately _did not
check_) the exit code of `elm-test --help` in light of
https://github.com/rtfeldman/node-test-runner/issues/359.

A prime example of why static typing is a boon for programming languages.

### DIFF
--- a/tests/flags.js
+++ b/tests/flags.js
@@ -132,6 +132,7 @@ describe('flags', () => {
 
       const runResult = execElmTest(['install', 'elm/regex']);
 
+      assert.ok(Number.isInteger(runResult.status));
       assert.notEqual(runResult.status, 0);
     }).timeout(60000);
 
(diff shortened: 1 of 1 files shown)


harrysarson:
See this screenshot showcasing a) the vscode debugger b) the object returned by `spawn.sync()` which has property `status` but not `code` and c) my artistic ability in mspaint.

image

harrysarson:
I added 65fd43bc94233ff082b61442e0cf7fe5ee6a18f7 which ensures the test values will remain integers going forwards.

harrysarson:
This has been sitting for a month and has a ✔️review so I am gonna merge. Hope that is okay!

PR #378, opened 2019-08-11
use elm-analyse to tidy elm code

harrysarson:
The first commit removes dead code, unused imports and renames all
unused variables to `_`.

In the second I ran `elm-format` on the elm files in `/elm/`. I changed the `npm run prettier:write/check` scripts to `npm run format:write/check` which now write and check both js files (with prettier) and elm files (with elm-format).

~Finally I updated the elm compiler to the no deps version!~


harrysarson:
AppVeyor failure is due to this: https://github.com/avh4/binwrap/issues/35

harrysarson:
I have no idea why using elm@0.19.0-no-deps causes test failure on windows. :(

avh4:
> I have no idea why using elm@0.19.0-no-deps causes test failure on windows

yeah, I ran into that on a previous PR somewhere... I think the failure then was related to concurrent execution, so maybe the no-deps wrapper is not safe to run concurrently on windows?

harrysarson:
> yeah, I ran into that on a previous PR somewhere... I think the failure then was related to concurrent execution, so maybe the no-deps wrapper is not safe to run concurrently on windows?

Makes sense, I wonder if we could insert a `sleep` or something to remove the concurrent execution in CI?

harrysarson:
Bah, I give up. This is ready to merge from my end (probably want to squash it).

harrysarson:
@rtfeldman what are your thoughts on adding elm-format to our workflow? Is this a dependancy and CI check you are happy to add?

harrysarson:
bump 🙂

PR #379, opened 2019-08-13
Upgrade version of node-elm-compiler

jwoudenberg:
`node-elm-compiler` recently got a new patch version to address a security vulnerability in `lodash`. This upgrades `node-test-runner` to use this new version, to address the security vulnerability here too.

Addresses #360

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "lodash": "4.17.15",
     "minimist": "^1.2.0",
     "murmur-hash-js": "1.0.0",
-    "node-elm-compiler": "5.0.3",
+    "node-elm-compiler": "5.0.4",
     "split": "1.0.1",
     "supports-color": "7.0.0",
     "temp": "0.9.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -496,23 +496,18 @@
       }
     },
     "find-elm-dependencies": {
-      "version": "2.0.1",
-      "resolved": "https://registry.npmjs.org/find-elm-dependencies/-/find-elm-dependencies-2.0.1.tgz",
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -42,7 +42,7 @@
     "lodash": "4.17.15",
     "minimist": "^1.2.0",
     "murmur-hash-js": "1.0.0",
-    "node-elm-compiler": "5.0.3",
+    "node-elm-compiler": "5.0.4",
```
smucode: Doesn't it make sense to make it `~5.0.4` so we get patch updates without a new release?

jwoudenberg: I think the `package-lock.json` file will pin it down anyway. I think that's intentional, for applications to pin dependencies so they don't accidentally break when some dependency does a patch update that turns out not be a patch update. Don't have a super strong opinion though and just kept this the way it was. Happy to change it!

smucode: If I understand https://docs.npmjs.com/files/package-lock.json it seems like it will be ignored: 

> One key detail about package-lock.json is that it cannot be published, and it will be ignored if found in any place other than the toplevel package.

So if you keep a package-lock.json in your project, it will ensure things don't accidentally get updated unless you want it to.

jwoudenberg: I did not know that, thanks for sharing!

I'm not sure what to do here. Almost all packages in the package.json are currently pinned, so I feel maybe this is a broader question? I'm currently inclined to leave it pinned on this PR, and maybe we can create a separate PR/issue to propose using `~` for all dependencies?

jwoudenberg:
@harrysarson this PR contains a change to `package.json` too, which is the change we're really after. Would restricting the PR to just that change address your concerns?

harrysarson:
The first commit looks good 👍 it is only the second (which does _not_ change `package.json` that concerns me). So yes!

jwoudenberg:
Excellent! I've force-pushed to the branch to remove the offending commit. Could you let me know if the updated PR looks good to you?

PR #381, opened 2019-09-17
Update chokidar, flow-bin, xml2js and cross-spawn

harrysarson:
Is good to keep our dependancies as uptodate as possible so here are some bumps.

The major version change in cross-spawn is to drop support for node < 8; the test runner only supports node >= 8 so this is not an issue.


harrysarson:
Sorry I did not mean to close, had issues rebasing!

harrysarson:
I will merge on friday if no one objects before then :)

PR #386, opened 2019-10-04
Update chokidar, which and flow-bin

harrysarson:
Another set of bumps. Chokidar looks like it is still having trouble with windows.

It has occured to me that we do not have a test for a custom elm install location, this would be good to add before bumping which.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -209,9 +209,9 @@
       }
     },
     "chokidar": {
-      "version": "3.2.0",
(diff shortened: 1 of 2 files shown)


rtfeldman:
@harrysarson do you feel comfortable merging this without the test?

If not, would probably be good to get the chokidar bump in separately.

harrysarson:
I got half way through writing a test, but that was enough to have manually confirmed that which 2.0.1 works :)

PR #394, opened 2019-11-07
A whole load of bumps

harrysarson:
Checkout the commit messages for more details - there looks to be nothing to spicy here though. 

I will merge friday week (2019/11/15) if no one objects before then :slightly_smiling_face:


harrysarson:
I have push some more upgrades, there are some MAJOR changes here but nothing that concerns us (mainly dropping support for node versions that we do not support).

rtfeldman:
Seems good - after you merge this, I can do a patch release over the weekend with this plus @stoeffel's `elmi-to-json` upgrades.

PR #397, opened 2019-11-11
Run test suite twice due to flaky tests

harrysarson:
We have a couple of flaky tests e.g.

```
  1) flags
       --fuzz
         Should use the provided value:
     AssertionError [ERR_ASSERTION]: '5' == undefined
      at Context.<anonymous> (tests\flags.js:260:14)
      at processImmediate (internal/timers.js:439:21)
```

causing valid pull requests to go red.

This is a hack, and a nasty one at that but it will hopefully mean less CI runs need to be manually restarted. It would be good to find and fix the root cause too though

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,4 +16,4 @@ install:
   - npm install
 
 script:
-  - npm test
+  - npm test || npm test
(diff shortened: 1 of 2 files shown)


harrysarson:
This doesn't work alas

PR #398, opened 2019-11-11
include debug info in assertion error

harrysarson:
For a flaky test that does not fail on my computer
but does (often) fail in CI print extra info
in the error message to help with debugging.

Should be reverted later (or never merged)

### DIFF
--- a/tests/flags.js
+++ b/tests/flags.js
@@ -234,7 +234,7 @@ describe('flags', () => {
       ]);
       const firstOutput = JSON.parse(runResult.stdout.split('\n')[0]);
 
-      assert.equal('12345', firstOutput.initialSeed);
+      assert.equal('12345', firstOutput.initialSeed, runResult.stdout);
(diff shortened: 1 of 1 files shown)


rtfeldman:
Computers. 🤕 

I'd be ok with adding a comment above it explaining why we have the third argument, and then merging.

harrysarson:
I was hoping that the CI on this pull requests would fail, but infuriatingly it hasn't. Will re-run and see!

harrysarson:
I got the error:

Somehow `$ elm-test tests/Passings/One.elm --report=json` prints only

    {"event":"runComplete","passed":"1","failed":"1","duration":"100","autoFail":null}

That should be the last line of stdout - there should be other lines first. Still cannot reproduce on my (linux) machine. How odd.

harrysarson:
Don't think this test is flaky anymore. Either way this PR is going nowhere.

PR #404, opened 2020-01-07
Fix #391

### LINKED ISSUE #391: Targeting specific tests not working consistently
On the current release (`0.19.1`), pointing to specific paths to look for tests does not work consistently. This seems to be a caching issue.

There is some info on how to reproduce this over here: https://github.com/zwilias/elm-demystify-decoders/issues/9

Deleting `elm-stuff/generated-code` between runs fixes the issue. Deleting `elm-stuff/generated-code/[...]/elm.json` _partially_ fixes the issue: new tests are added to the set, but old tests run twice!

@rtfeldman this might interest you 😄

jwoudenberg:
Paired with @rtfeldman

### DIFF
--- a/lib/Runner.js
+++ b/lib/Runner.js
@@ -103,6 +103,8 @@ function findTests(
   return new Promise(function(resolve, reject) {
     function finish() {
       const args = elmVersion ? ['--elm-version=' + elmVersion] : [];
+      // TODO: Have elmi-to-json only read test file paths, not all paths.
+      // args = args.concat(testFilePaths);
       let proc = spawn(readElmiPath, args.concat(['--for-elm-test']), {
         cwd: testRootDir,
         env: process.env,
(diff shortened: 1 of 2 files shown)


lib/Runner.js
```diff
@@ -103,6 +103,8 @@ function findTests(
   return new Promise(function(resolve, reject) {
     function finish() {
       const args = elmVersion ? ['--elm-version=' + elmVersion] : [];
+      // TODO: Have elmi-to-json only read test file paths, not all paths.
+      // args = args.concat(testFilePaths);
```
rtfeldman: More context on this: it would be more efficient to have `elmi-to-json` only read the `.elmi` files that will be used.

However, we couldn't determine how to do that, and didn't want to block the fix on that. Instead, we're filtering them out after the fact below.

terezka:
I wonder if we should add a test to prevent another regression? Don't have context enough to know if that makes sense, so let me know!

jwoudenberg:
I think it does make sense, and we sort-of did a thing like that, in that we modified an existing test to reproduce this problem. The relevant line that enabled this was this:

https://github.com/rtfeldman/node-test-runner/pull/404/files#diff-669efeda79bb547b3537ab544fd28f75R149

terezka:
Oh, didn't realize that was what that was! Great!

zwilias:
<3

PR #410, opened 2020-02-25
Bumps

harrysarson:
The MAJOR change in xmlbuilder is to drop support for node 6. We no longer support node 6 so this is not a problem.

Otherwise we have minor changes to elm/core and flow-bin


harrysarson:
Added a flow-bin bug fix to the batch.

PR #418, opened 2020-04-09
only run prettier on lts node

harrysarson:
a) this means folk can see test failures and lint failures seperately
b) We can upgrade to prettier@2 which refuses to run on node 8 (boo).

(b) is more important here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,10 +10,13 @@ os:
   - linux
   - osx
 
-install:
-  - node --version
-  - npm --version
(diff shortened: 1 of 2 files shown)


harrysarson:
@rtfeldman. @mgold, @avh4 

I would appreciate some eyes on this one. The alternative is to stick with prettier@1 until the next breaking release of the test runner.

harrysarson:
Thanks! The prettier the better that's what I say?

PR #422, opened 2020-05-21
Separate paths with newlines

Arkham:
Hi there! I noticed that when passing a very long list of files Elm 0.19.1 will blow up and report 

```
[2020-05-21T10:49:04.768Z] -- UNFINISHED LIST ----------------------- src/Test/Generated/Main2355714466.elm
[2020-05-21T10:49:04.768Z] 
[2020-05-21T10:49:04.768Z] I cannot find the end of this list:
```

Opening the file locally led me to discover that the incriminated line had around 74'000 characters. So in this PR I've separated each path with a newline, let me know if this seems the right approach!

### DIFF
--- a/lib/Generate.js
+++ b/lib/Generate.js
@@ -396,7 +396,7 @@ function generateMainModule(
     fuzz: isNaN(fuzz) ? 'Nothing' : 'Just ' + fuzz,
     seed: seed,
     report: getReportCode(),
-    paths: testFilePaths.map(sanitizedToString).join(',\n      '),
+    paths: testFilePaths.map(sanitizedToString).join(','),
   };
 
   var optsCode =
--- a/lib/Generate.js
+++ b/lib/Generate.js
@@ -396,7 +396,7 @@ function generateMainModule(
     fuzz: isNaN(fuzz) ? 'Nothing' : 'Just ' + fuzz,
     seed: seed,
     report: getReportCode(),
-    paths: testFilePaths.map(sanitizedToString).join(','),
+    paths: testFilePaths.map(sanitizedToString).join(',\n      '),
   };
 
   var optsCode =
--- a/example-application-lots-of-files/.gitignore
+++ b/example-application-lots-of-files/.gitignore
@@ -1,0 +1,2 @@
+/elm-stuff
+/tests/TestsPassing*
--- a/example-application-lots-of-files/create_tests.sh
+++ b/example-application-lots-of-files/create_tests.sh
@@ -1,0 +1,30 @@
+#! /bin/bash
+
+set -euxo pipefail
+
+TEMPLATE=$(cat <<-EOS
+
+import Expect
+import Something
+import Test exposing (Test, test)
+
+
+testEqual : Test
+testEqual =
+    test "Expect.equal works" <|
+        \\() ->
+            Something.ultimateAnswer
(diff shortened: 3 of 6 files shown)


harrysarson:
CI is flaky so do not worry about appveyor.

Could this cause issues where the compiler does not like the indentation?

Arkham:
By looking at the file we generate right now, I feel it's already not formatted as elm-format would generate it. Would it be possible to run elm-format on the file after we generate it? In this way, we can use it to double check that everything is ready to run :)

harrysarson:
I think that would slow test runs down with out much gain. The generated file doesn't have to be pretty along as the elm compiler can parse it.

Arkham:
Ah, good point! I think Elm should be able to parse the file fine, but I haven't been able to use this branch to repro. Sure, here it is: https://gist.github.com/Arkham/a354b4de401a38ce33d32196ddc31fe7

lib/Generate.js
```diff
@@ -396,7 +396,7 @@ function generateMainModule(
     fuzz: isNaN(fuzz) ? 'Nothing' : 'Just ' + fuzz,
     seed: seed,
     report: getReportCode(),
-    paths: testFilePaths.map(sanitizedToString).join(','),
+    paths: testFilePaths.map(sanitizedToString).join(',\n      '),
```
harrysarson, line 399:
```suggestion
    paths: testFilePaths.map(sanitizedToString).join(',\n          '),
```

Looking at the generated file I think 8 spaces is best. (I think the elm compile only complains about too little indentation so adding extra is safe)

harrysarson:
@Arkham thanks! Would you be able to add an integration test that fails without this change? If you feel that adding the test is too much then I am happy to merge this as is but more tests are always nice :) 

---

For the integration test:

Add a new directory `example-application-lots-of-files` and then some thing like:

```
shell.echo('### Testing elm-test on example-application-lots-of-files/');

shell.cd('example-application-lots-of-files');

assertTestSuccess(???, false);

shell.cd('../');
```

Where ??? depends on the command line args you passed to elm-test to get the error.

Arkham:
I'll add a test today, thanks for the tips!

harrysarson:
Hey @Arkham could you confirm the status on this one. Before merging we would need to fix tests on windows (probably by porting the bash script to javascript). I would be happy to do that once we are happy with everything else.

It looks like the last commit reverts the fix to try and demonstrate the new test failing. However, the new test still passes (on linux atleast).

Arkham:
@harrysarson that was quite puzzling to me as well! It feels like some more debugging needs to be done to figure out the actual cause of the error (since I can sometimes reproduce it locally), but I haven't been able to spend time on this. Will report back when I find out more :)

harrysarson:
Makes sense, take your time!

harrysarson:
Hey @Arkham, not wanting to cause any stress by pinging (please feel free to ignore). I am just curious to hear if you have found any thing interesting?

Arkham:
Hey Harry, I haven't encountered this issue in our builds any longer, so I haven't spent much time investigating this :) I'd be happy to close this issue and reopen when/if I have a more reproducible path to an error. Does that sound good to you?

harrysarson:
That sounds good to me! Thanks for your investigation here and do let us know if the issue resurfaces.

PR #425, opened 2020-05-30
Remove unused stuff, improve performance and make elm-test more light-weight

lydell:
- There’s a bunch of unused code in this repo. I added ESLint to find it all and get rid of it.

- Removed `filterExposing`. That function wasn’t doing anything <del>since the upgrade to <code>elmi-to-json --for-elm-test</code></del> <ins>since the upgrade to Elm 0.19</ins>. This means we no longer need to spend time on parsing the top of each elm file ourselves! (Edit: Elm 0.18 includes unexposed top-level values in .elmi files, so elm-test had to filter the unexposed ones out. Elm 0.19 only includes _exposed_ values in .elmi files, so there’s no need to filter anymore.)

- Fixed an oversight where `elmi-to-json` was called twice in a row.

- Fixed the Node version check to align with `"engines"` in package.json.

- Got rid of lodash. Paired with https://github.com/avh4/binwrap/pull/45, elm-test is roughly 5 MB lighter!


.eslintignore
```diff
@@ -0,0 +1,3 @@
+elm-stuff
+fixtures
+templates
```
harrysarson, line 3: Does this have to be its own file or could we include in package.json? I have a slight preference for less file in the top level directory but no stress if this is the only way to exclude files?

.travis.yml
```diff
@@ -2,9 +2,9 @@ sudo: false
 
 language: node_js
 node_js:
-    - "8"
-    - "10"
-    - "12"
+  - '8'
+  - '10'
+  - '12'
```
harrysarson, line 7: While we are at it would it be possible to add node 14 to the list?

appveyor.yml
```diff
@@ -1,9 +1,9 @@
 environment:
-  ELM_VERSION: "0.19.1"
+  ELM_VERSION: '0.19.1'
   matrix:
-  - nodejs_version: "12"
-  - nodejs_version: "10"
-  - nodejs_version: "8"
+    - nodejs_version: '12'
+    - nodejs_version: '10'
+    - nodejs_version: '8'
```
harrysarson, line 6: Ditto for node 14

lib/Runner.js
```diff
@@ -86,7 +76,11 @@ function moduleFromFilePath(filePathArg) {
   // e.g. 'tests/All/Passing.elm' ===> ['Passing', 'All', 'tests']
   // This way, if we're given 'All.Passing' as a module name, we can also
   // flip it into ['Passing', 'All'], and see if the first N elements line up.
-  return _.compact(parsed.dir.split(path.sep).concat([basename])).reverse();
+  return parsed.dir
+    .split(path.sep)
+    .concat([basename])
+    .filter(Boolean)
```
harrysarson, line 82: Out of curiousity: what does `.filter(Boolean)` do here?

lib/Runner.js
```diff
@@ -96,80 +90,73 @@ function moduleFromTestName(testName) {
...
-              .then(resolve)
-              .catch(reject);
-          })
-          .catch(reject);
-      });
-    }
-
-    return finish();
+        })
+        .filter(Boolean);
```
harrysarson, line 144: Could extract this into a `filterMap` function for clarity?

lib/Supervisor.js
```diff
@@ -216,10 +216,10 @@ function run(
     });
 
     server.on('listening', function () {
-      workers = _.range(0, processes).map(function (index) {
+      for (var i = 0; i < processes; i++) {
```
harrysarson, line 219: I do like the functional style of the old approach. Do you think we could use something like this?

```js
workers = Array.from({length: process}).map((_, index) => {
```

lib/elm-test.js
```diff
@@ -450,15 +448,9 @@ function generateAndRun(
...
     )
-      .then(function () {
-        return Runner.findTests(
-          elmVersion,
-          generatedCodeDir,
-          testFilePaths,
-          sourceDirs,
-          !Compile.isMachineReadableReporter(report)
-        )
-          .then(function (testModules) {
```
harrysarson, line 461: What was this function call doing before?

harrysarson: Ah I should have read the description first!

> - Fixed an oversight where elmi-to-json was called twice in a row.

I guess this was what it was doing? Nice catch :)

harrysarson:
So we have a flaky test that almost always causes one of the appveyor runs to fail (travis runs the suite twice if it fails the first time).

I am not seeing any appveyor failures which means that flaky test has surely been fixed! I am doing some more testing of CI in #426 just to confirm this isn't too good to be true). Fixing this flakiness gets a huge :+1: from me.

lib/Runner.js
lydell: `Boolean` has the type `a -> Bool`. It turns truthy things into `true`, and falsy things into `false`. So the effect of `.filter(Boolean)` is that you only keep truthy things. Flow also has a special case for `.filter(Boolean)` that actually changes the type of the array (so you can get rid of `null`s for example: https://github.com/facebook/flow/blob/b9f3f436905537817f64e129f3428f612aba6683/lib/core.js#L225

lib/Supervisor.js
lydell: Why complicate it though?

.eslintignore
lydell: Feel free to change it if you like.

lib/Runner.js
lydell: I switched to explicitly filtering out empty strings.

lib/elm-test.js
lydell: Correct!

.eslintignore
lydell: Personally I like it being easy to copy-paste between .gitignore, .eslintignore, .prettierignore, .dockerignore

.travis.yml
```diff
@@ -13,7 +14,6 @@ os:
 script:
   - travis_retry npm run test-only
```
harrysarson, line 15: This PR has fixed our flakiness in CI!

```suggestion
  - npm run test-only
```

.eslintignore
harrysarson: Okay that makes sense. Maybe at some point in the future we can consider merging our config files but this is good for now :)

harrysarson:
I have one more request: could you summarise what would brake if we remove `filterExposing` without using `elmi-to-json --for-elm-test`? (It would be even better if you we could add a test that ensures we don't regress here if elmi-to-json changes etc. Definitely does not need to be part of this PR though).

Otherwise, this looks really good to me. I think I will leave it open for ~2 weeks as it is quite a large change to a very fragile codebase and it would be good to get some more eyes on it. After that though then we should merge (even if we don't get any more reviews) because we definitely want the flaky tests fixed. I will also make sure that no other pull requests land until after this to save us any conflicts.

How does that sound? And do please ping in 2 weeks if I forget :)

lydell:
> could you summarise what would brake if we remove filterExposing without using `elmi-to-json --for-elm-test`?

I don’t know the answer to that question, but I could try to dig it up during the coming two weeks! It might not even be related to the `--elmi-to-json` flag – it could also be a left-over from a previous Elm version or something.

lydell:
Some findings so far:

`filterExposing` comes from https://github.com/rtfeldman/node-test-runner/commit/657100d1de304ca1e4dd0f356204404463822a2c (2017), part of https://github.com/rtfeldman/node-test-runner/pull/107

As far as I can tell, elm-test was using https://github.com/stoeffel/elm-interface-to-json back then.

Maybe Elm 0.18 included _all_ toplevel values in its .elmi files? If so, `filterExposing` should ideally have been removed when upgrading to 0.19, but since it didn’t break anything (it only became useless) it’s of course an easy oversight to make. Maybe I’ll try installing Elm 0.18 and elm-interface-to-json and see what I find…

Regarding a test – that should be easy, I think. I’ll just need to make a file where _some_ top-level values are tests but not exposed. The unexposed tests should be failing, so that if they _were_ picked up the test would fail.

harrysarson:
That all makes sense, and the test sounds great! It is sad that elmi no longer has non exposed things in as it may mean folks tests are going undetected and therefore bugs are slipping through the net. Nothing we can do about it in the test runner though. :man_shrugging:

lydell:
In Slack I learned that some people actually _want_ unexposed test values though, to group them under a common description:

```elm
all =
    describe "..."
        [ someTest
        , otherTest
        ]
someTest =
    ...
otherTest =
    ...
```

So a checker for undetected tests needs to be more advanced: It needs to check for test values that not only unexposed, but also unused.

lydell:
Confirmed: Elm 0.18 .elmi files contain unexposed top-level values. https://github.com/lydell/elm-0.18-elmi-test

lydell:
Test added – let me know if you think it’s good.

tests/fixtures/tests/Passing/Unexposed.elm
```diff
@@ -0,0 +1,30 @@
+module Passing.Unexposed exposing (all)
+
```
harrysarson, line 2:
```suggestion

{- Ideally we would detect unexposed tests and warn users about them. In elm 0.18.0 we did this, but 
internal compiler changes mean we no longer can. This test passes if unexposed tests are ignored to
keep the behavior consistent. Ideas about how we can detect unexposed tests (and thus cause this 
test to fail) are very welcome!

See https://github.com/rtfeldman/node-test-runner/pull/425#issuecomment-637028958 (and following 
comments) for more info.
}
```

tests/fixtures/tests/Passing/Unexposed.elm
```diff
@@ -0,0 +1,30 @@
...
+testTrue =
+    test "Expect.true works" <|
+        \() ->
+            True
+                |> Expect.true "this should never fail!"
+
+
+testUnexposed : Test
+testUnexposed =
+    test "This test is unexposed and should not run!" <|
```
harrysarson, line 28:
```suggestion
    test "This test is unexposed and will not run!" <|
```

harrysarson:
Thanks @lydell this is really nice work on this PR. I added some info into the test, if you are happy with it I will click the merge button!

tests/fixtures/tests/Passing/Unexposed.elm
```diff
@@ -0,0 +1,38 @@
+module Passing.Unexposed exposing (all)
+
+{- Ideally we would detect unexposed tests and warn users about them. In elm 0.18.0 we did this, but 
+internal compiler changes mean we no longer can. This test passes if unexposed tests are ignored to
+keep the behavior consistent. Ideas about how we can detect unexposed tests (and thus cause this 
+test to fail) are very welcome!
+
+See https://github.com/rtfeldman/node-test-runner/pull/425#issuecomment-637028958 (and following 
+comments) for more info.
+}
```
harrysarson, line 10:
```suggestion
-}

```

Issue #431, opened 2020-07-28
Trim console output


andrewMacmurray:
(Based on this discussion https://discourse.elm-lang.org/t/elm-test-output/6084)

Something I've run into as the project I'm working on gets larger is the console output's file list can grow very large

```
elm-test 0.19.1
---------------

Running 80 tests. To reproduce these results, run: elm-test --fuzz 100 --seed 231172734506082 
/Users/andrewmacmurray/oprah/application/frontend/src/app/elm/Account.elm 
/Users/andrewmacmurray/oprah/application/frontend/src/app/elm/Account/AccessLevel.elm 
/Users/andrewmacmurray/oprah/application/frontend/src/app/elm/Account/Admin.elm 
/Users/andrewmacmurray/oprah/application/frontend/src/app/elm/Account/AllowedModule.elm 
/Users/andrewmacmurray/oprah/application/frontend/src/app/elm/Account/Partner.elm
...400-500 lines

TEST RUN PASSED

Duration: 526 ms
Passed:   80
Failed:   0 
```

For my particular case I don't really benefit from seeing all the file paths - but other may find this useful.

For me it would be really nice to see the output like:

```
elm-test 0.19.1
---------------

Running 80 tests. To reproduce these results, run: elm-test --fuzz 100 --seed 231172734506082 

TEST RUN PASSED

Duration: 526 ms
Passed:   80
Failed:   0 
```

One idea could be to pass an additional flag, I'm not wild about these names but maybe something like:

+ `--trim`
+ `--trim-output`
+ `--hide-paths`

```
elm test --report=console --trim
```

Or if it was on by default maybe there could be a `--verbose` flag

Another option could be have a different reporter, something like:

```
elm test --report=compact
```

Any thoughts? Very happy to make an PR if we arrive at a conclusion for this 😄

harrysarson:
Are the 400-500 lines needed if you wanted to reproduce the test result?

gampleman:
I wonder if this couldn't be alleviated if relative paths were used? (We could even output the cwd if we wanted to)

```
elm-test 0.19.1
---------------

Running 80 tests. To reproduce these results, run: 

cd /Users/andrewmacmurray/oprah/application/frontend/src/app/elm
elm-test --fuzz 100 --seed 231172734506082 Account.elm Account/AccessLevel.elm Account/Admin.elm Account/AllowedModule.elm Account/Partner.elm
...

TEST RUN PASSED

Duration: 526 ms
[...]

PR #432, opened 2020-08-06
Trim console reporter output (#431)

marc136:
Hello, this is my take on reducing the console output to a more manageable amount as discussed in #431 

Now it returns a short reproduction command like:
`Running 407 tests. To reproduce these results, run: elm-test --fuzz 100 --seed 115440003747078`
or
`Running 224 tests. To reproduce these results, run: elm-test --fuzz 100 --seed 41574277908519 tests\One.elm tests\Group\*`
instead of printing many many lines of absolute file paths.

---

I changed the data that is passed to all the reporters. I'm confident that it does not break anything for the console reporter, as it only printed the list of files before this commit, and now it prints the list of file globs passed to elm-test.

I'm not sure if it might break something for another tool that works with the output of the json reporter (e.g. if it needs the list of absolute paths of every found test file).

I would really appreciate if this (or a similar change) would make it into this nice tool.

### DIFF
--- a/elm/src/Test/Reporter/Console.elm
+++ b/elm/src/Test/Reporter/Console.elm
@@ -92,8 +92,8 @@ textToValue useColor txt =
         |> Encode.string
 
 
-reportBegin : UseColor -> { paths : List String, fuzzRuns : Int, testCount : Int, initialSeed : Int } -> Maybe Value
-reportBegin useColor { paths, fuzzRuns, testCount, initialSeed } =
+reportBegin : UseColor -> { r | globs : List String, fuzzRuns : Int, testCount : Int, initialSeed : Int } -> Maybe Value
+reportBegin useColor { globs, fuzzRuns, testCount, initialSeed } =
     let
         prefix =
             "Running "
@@ -103,7 +103,7 @@ reportBegin useColor { paths, fuzzRuns, testCount, initialSeed } =
                 ++ " --seed "
                 ++ String.fromInt initialSeed
     in
-    (String.join " " (prefix :: paths) ++ "\n")
+    (String.join " " (prefix :: globs) ++ "\n")
         |> plain
         |> textToValue useColor
         |> Just
--- a/elm/src/Test/Reporter/JUnit.elm
+++ b/elm/src/Test/Reporter/JUnit.elm
@@ -5,7 +5,7 @@ import Test.Reporter.TestResults exposing (Failure, Outcome(..), SummaryInfo, Te
 import Test.Runner.Failure exposing (InvalidReason(..), Reason(..))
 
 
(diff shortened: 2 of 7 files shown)


harrysarson:
Thanks  @marc136, this looks great! However, I am uneasy about the changes to the json and junit reporter though. I am unsure exactly what stability guarantees we make but I think this change to the json/junit output could break things for people.

My thoughts are this:

1. Keep threading the "paths" value in the same way we currently do (i.e. revert changes to [this line](https://github.com/rtfeldman/node-test-runner/pull/432/files#diff-047f238be7300455dd55a8c22225e53fL460)) but **also** thread through a new value called `globs`. Then we can use the `globs` value in the console reporter and keep using the `paths` value in the json. [1]

2. Delay this change until the next major release of the test runner (i.e. when elm 0.19.2 gets released).

What do you think?

---

[1]: even better would be to include the `globs` value **aswell** as the `paths` value in the json we output as I think the globs is probably more useful. Moreover, including both would let us also to option (2) at a later date.

marc136:
Hello @harrysarson,

for my change, I started from the test reporters and then followed the variables up to the point where I made the changes.
The paths value is only used insided `reportBegin` of all reporters ([link to code](https://github.com/rtfeldman/node-test-runner/blob/8465763f223e91239e2d5fc4b785942ac823b20f/elm/src/Test/Reporter/Reporter.elm#L19))
The [junit reporter does not use the paths value](https://github.com/rtfeldman/node-test-runner/blob/8465763f223e91239e2d5fc4b785942ac823b20f/elm/src/Test/Reporter/JUnit.elm#L10) and the [json reporter prints it](https://github.com/rtfeldman/node-test-runner/blob/8465763f223e91239e2d5fc4b785942ac823b20f/elm/src/Test/Reporter/Json.elm#L14).

Regarding 1 (and [1]):
I like that proposal, and I think the change could come in any future revision. So this would mean:
- The `globs` value is threaded along through all parts (together with `paths`) and added to the `RunInfo` record
- `Test.Reporter.Console.reportBegin` prints the globs in the reproduction command (instead of all files)
- `Test.Reporter.Json.reportBegin` prints the globs as an additional json field
- `Test.Reporter.Junit.reportBegin` is not changed and keeps returning `Nothing`

Regarding 2:
I think it would be good to mark the `paths` value as deprecated for consumers of the json reporter, but I saw no obvious way to do that and then remove it with the next major release.
We could ask in discourse and slack if anyone is aware of a tool that uses the `paths` value of the json reporter.

Maybe we can create a new issue for that like "Remove the `paths` value from the JSON reporter" and list all known tools that use the json report so they could be checked before the change?

harrysarson:
Sounds like a good plan!

marc136:
I changed the PR to act as described in my last commit, @harrysarson.

harrysarson:
Thanks @marc136!

Issue #424, opened 2020-05-30
Why are npm dependencies pinned?


lydell:
Almost all npm dependencies of this package are pinned:

https://github.com/rtfeldman/node-test-runner/blob/b86ed4f87ee6a5908392afa5afdf4bc5443c5f38/package.json#L35-L51

Why? Isn’t it a good thing to allow a range, so people can get bug fixes and “security” fixes (which `npm audit` nags about) without having to update elm-test?

harrysarson:
Thanks!

PR #435, opened 2020-08-26
Reduce number of external deps

harrysarson:
I am working up to a new release. My thinking is `less deps == less things that can go wrong` and also `less deps == less time waiting for upstream packages to fix`. 

I also want to unpin all our dependencies so that upstream patches can benefit elm-test users without us having to do an update. Unpinning the dependencies does make us more vulnerable to someone nasty adding a patch to an npm package to mine bitcoin (it has [happened before](https://snyk.io/blog/malicious-code-found-in-npm-package-event-stream/)) which is another good reason to have less deps.

It is probably easiest to review commit by commit and with the github ignore whitespace option switched on!

### drop dep on byline
    
    We can, as far as I can tell, use node's standard readline module
    instead.

### drop dep on temp
    
    temp is unmaintained (https://github.com/bruce/node-temp/issues/67).
    Instead of using tempory directories we
    
      as part of elm-test:  create a tempory directory within elm-stuff.
      within our testsuite: create a tempory directory within the
                            fixtures directory. The directory should be
                            deleted after each test runs and, to be on
                            the safe side, I have added the directory
                            to our gitignore.

### drop dependancy on supports-color
    
    Chalk exposes this value so use chalk rather than supports-color
    directly (https://github.com/chalk/chalk#chalksupportscolor). We
    still depend transitively on supports-color but atleast we do not
    have to keep track of the versioning ourselves.

### drop dependancy on firstline
    
    The logic is straightforward so I inlined it into Runner.js. Either node 8 does
    not support async/await or we have configured eslint wrongly. Eitherway, I have
    avoided using async/await for now. We can asyncify the codebase (if we want)
    after dropping node 8 support.

### drop dependancy on find-parent-diff
    
    Whilst this may look like it breaks elm-test when we run from a subdirectory
    of the current project, it does not. elm-test is already broken when run
    from a subdirectory. This just removes code that is broken and does not work.
    If anyone wants to fix this reverting this commit would not be a bad place to
    start.
    
    In the meantime, one less dependancy is a good thing. All the more so becase
    https://github.com/thlorenz/find-parent-dir has not been updated for 7 years.


harrysarson:
I realise this is a bit cheeky but @lydell you are the last person to do any major coding in this repository and also have an interest in seeing a new version of elm-test released (so folk can benefit from #425).

As I am the only person reviewing PR's here. These commits are unlikely to get any eyes on them. If you happened to have the time and the inclination, I would hugely appreciate a review from you! Absolutely no pressure 😃

harrysarson:
CI failure is spurious.

lib/Runner.js
```diff
@@ -1,9 +1,33 @@
...
+    });
+    reader.on('close', () => {
+      if (!foundLine) {
+        reject(new Error(`File ${pathToFile} is empty!`));
+      }
+    });
+    reader.on('error', (err) => reject(err));
+  });
+}
+
```
avh4: Would it be worth sticking this in a separate file just to keep clutter out of this file?

lib/Install.js
```diff
@@ -1,115 +1,124 @@
...
+  const dirPath = path.join(
+    Compiler.getGeneratedCodeDir(process.cwd()),
+    'install'
+  );
 
   try {
-    var dirPath = temp.mkdirSync('elm-test-install-');
+    // Recreate the directory to remove any artifacts from the last time
+    // someone ran `elm-test install`. We do not delete this directory after
+    // the installation finishes in case the user needs to debug the test run.
```
avh4, line 19: :+1:

tests/flags.js
```diff
@@ -7,7 +7,7 @@ const spawn = require('cross-spawn');
 const fs = require('fs-extra');
 const os = require('os');
 const xml2js = require('xml2js');
-const byline = require('byline');
+const readline = require('readline');
```
avh4, line 10: oh, byline was only used in tests?

lydell:
I’ll try to have a look later today. All I can say for now is that we can reduce the number of  dependencies even more via https://github.com/avh4/binwrap/pull/45 (elmi-to-json depends on binwrap, but unfortunately the version is [pinned](https://github.com/stoeffel/elmi-to-json/blob/ae40d1aa1e3d6878f2af514e611d44890e7abc1e/package.json#L30) so a PR would be needed there too…)

avh4:
thanks, just bumped up https://github.com/avh4/binwrap/pull/45 higher in my to-do list :)

lydell:
Also, do we want to unpin the remaining dependencies? https://github.com/rtfeldman/node-test-runner/issues/424

harrysarson:
Yes! I will unpin all dependencies (merge this first) before the revision-4 release. That also means that folk should benefit from https://github.com/avh4/binwrap/pull/45 even if we release before that lands.

tests/flags.js
harrysarson: Does byline do anything that readline does not? Unless find in files has let me down this was the only use of byline.

lydell: byline was in devDependencies so it must have been used for tests only.

It’s unclear to me what byline is providing over readline. But if the tests pass we’re all good, I’d say. We’re not doing anything complicated in the tests anyway.

lydell (approved):
Looks good to me too. The module replacements look sensible. The code seems to work locally. Appveyor fails, though?

harrysarson:
Our test suite is flaky. It only fails on one node version and the test that fails is a known offender. Thank you @lydell! Brilliant

harrysarson:
Thanks everyone! Next step: unpin deps.

PR #437, opened 2020-09-02
unpin all dependancies

harrysarson:


harrysarson:
Does this match your expectations @lydell?

package.json
```diff
@@ -40,28 +40,28 @@
...
-    "prettier": "2.0.5",
-    "shelljs": "0.8.4",
-    "strip-ansi": "6.0.0",
-    "xml2js": "0.4.23"
+    "elm": "latest-0.19.1",
+    "elm-format": "^0.8.3",
+    "eslint": "^7.6.0",
+    "flow-bin": "^0.131.0",
+    "mocha": "^7.2.0",
+    "prettier": "^2.0.5",
```
lydell, line 62: Prettier recommends an exact version. package-lock.json usually takes care of that anyway, but if we ever remove and regenerate it we won’t accidentally cause formatting changes in all files. Up to you if you want to change it or not.

lydell (approved):
Yes, this is what I had in mind!

package.json
harrysarson: Very true, maybe I should pin all dev deps? Do bad dev-deps show up in `npm audit`?

harrysarson:
```suggestion
    "prettier": "2.0.5",
```

lydell: I usually pin _all_ my devDeps. Just like I pin all dependencies in an application. I think ranges only ever make sense for sub-dependencies.

`npm audit` checks what’s in your `package-lock.json`:

```
❯ npm audit
npm ERR! code EAUDITNOLOCK
npm ERR! audit Neither npm-shrinkwrap.json nor package-lock.json found: Cannot audit a project without a lockfile
npm ERR! audit Try creating one first with: npm i --package-lock-only

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/lydell/.npm/_logs/2020-09-10T10_37_20_134Z-debug.log
```

Your own devDeps are in your package-lock.json, but not your dependencies’ devDeps. So it works like it should: Only contributors of node-test-runner will ever get to know about vulnerabilities in devDeps.

harrysarson: brilliant, that is exactly what we want

package.json
```diff
@@ -40,28 +40,28 @@
...
+    "shelljs": "^0.8.4",
+    "strip-ansi": "^6.0.0",
+    "xml2js": "^0.4.23"
```
harrysarson, line 65:
```suggestion
    "elm": "0.19.1.3",
    "elm-format": "0.8.3",
    "eslint": "7.6.0",
    "flow-bin": "0.131.0",
    "mocha": "7.2.0",
    "prettier": "2.0.5",
    "shelljs": "0.8.4",
    "strip-ansi": "6.0.0",
    "xml2js": "0.4.23"
```

PR #438, opened 2020-09-10
fix --compiler flag

harrysarson:
By resolving _after_ we use which to locate binaries on PATH we can
actually use binary names (that are on PATH) as well as paths to
binaries.

This commit fixes a bug I introduced in https://github.com/rtfeldman/node-test-runner/pull/330


harrysarson:
Errors are to do with the fact that you cannot just execute a text file on windows.

Issue #385, opened 2019-10-01
requiring elmi-to-json is slow on Windows


showell:
Even if I just run `elm-test --help`, it take 3.5s on my relatively powerful Windows laptop to run that in a git bash shell.

I am not sure what I'm doing wrong, or if elm-test is just heavy like that.

If there isn't any easy way to speed this up, could it at least be documented as a known issue in the readme?

showell:
This line is particularly slow (930ms on a fast, healthy Windows laptop with a current version of node):

```
var readElmiPath = require('elmi-to-json').paths['elmi-to-json'];
```

Razzeee:
Updating `elmi-to-json` to the latest version 1.2.0 should presumably help with this https://github.com/stoeffel/elmi-to-json/releases

But I can't get the tests to pass after that.

Razzeee:
Wos done in https://github.com/rtfeldman/node-test-runner/commit/7ac6505cf78aab0caad4d885a14a05c84eae1133 and should be in the next release

rtfeldman:
@showell are you still seeing this in the current release, or can we close this?

showell:
My issues were never resolved.  There's a thread here: https://elmlang.slack.com/archives/C0CLGCMMF/p1573296606141400

We figured out that it was Windows-specific issues.

lydell:
@showell Would you mind trying out https://github.com/rtfeldman/node-test-runner/pull/442 and see if it’s faster for you?

1. Clone and checkout my branch
2. Run the Windows equivalent of `time ~/path/to/node-test-runner/bin/elm-test` in your project

showell:
@lydell Sorry I can't help--I haven't been active in Elm lately, due to my day job.

harrysarson:
I am happy to close this now. We no longer require elmi-to-json (on any OS which includes windows) so it cannot be slow. If anyone has any perf issues with the new approach (not yet released, will be version `0.19.1-revison5`) please open a new issue :)

Issue #416, opened 2020-04-09
BREAKING changes


harrysarson:
A list of breaking changes to go in the next release (`0.19.2` or `0.20.0`):

## Drop support for node 8

- Update fs-extra https://github.com/harrysarson/node-test-runner/pull/99
- Update chalk https://github.com/harrysarson/node-test-runner/pull/112
- Update mocha https://github.com/harrysarson/node-test-runner/pull/128

harrysarson:
I would love to replace `.then(...)` with async/await too.

harrysarson:
Fixes by #461

Issue #248, opened 2018-01-29
Non-determinism with --seed argument


mpizenberg:
Hi, I was trying to reproduce a failed test and noticed some non-determinism in elm-test with a given seed. I did nothing at all (no file opening, no save, literally nothing) between the first run (no seed, with an error) and the second run (with provided seed). Yet, the second run does not reproduce the failed test.

If it helps, my configuration is:
- OS: 64-bit Archlinux, kernel version 4.9.77-1-lts
- Processor: i5-5300U
- elm stack: elm-make 0.18, elm-test 0.18.9, elm-community/elm-test 4.2.0, elm-lang/core 5.1.1
- node v9.2.1
- project: opensolid/geometry@29d9b1e1

The runs mentionned previously:

```
$ elm-test --fuzz 1000                                                                                                                3683ms  lun. 29 janv. 2018 11:18:32 +08
Success! Compiled 0 modules.
Successfully generated /dev/null
Success! Compiled 1 module.
Successfully generated /home/matthieu/git/perso/elm/packages/opensolid/geometry/elm-stuff/generated-code/elm-community/elm-test/elmTestOutput.js

elm-test 0.18.9
---------------

Running 183 tests. To reproduce these results, run: elm-test --fuzz 1000 --seed 221291260

↓ LineSegment2d
✗ Intersection should be symmetric

Given (LineSegment2d (Point2d (0.000001,10),Point2d (0.000001,-10)),LineSegment2d (Point2d (0,10),Point2d (0.0013658365545529956,-10)))

    Just (Point2d (9.999999999999997e-7,9.985356959488797))
    ╷
    │ Expect.equal
    ╵
    Just (Point2d (0.000001,9.985356959488797))



TEST RUN FAILED

Duration: 5876 ms
Passed:   182
Failed:   1

$ elm-test --fuzz 1000 --seed 221291260                                                                                                 8.4s  lun. 29 janv. 2018 11:18:43 +08
Success! Compiled 0 modules.
Successfully generated /dev/null
Success! Compiled 1 module.
Successfully generated /home/matthieu/git/perso/elm/packages/opensolid/geometry/elm-stuff/generated-code/elm-community/elm-test/elmTestOutput.js

elm-test 0.18.9
---------------

Running 183 tests. To reproduce these results, run: elm-test --fuzz 1000 --seed 221291260


TEST RUN PASSED

Duration: 5856 ms
Passed:   183
Failed:   0
```

mgold:
[...]

Issue #211, opened 2017-09-04
Command line flags aren't validated


boxed:
```
elm-test --asdkjasdjaslkdjasjldkj
```
just runs elm-test. It should complain that the flag is invalid/unknown. I discovered this when running

```
elm-test --add-dependencies
```
and it didn't do anything. I expected it to update tests/elm-packages.json or complain about a missing argument.

harrysarson:
Would be great to check if elm-test still ignores invalid flags.

boxed:
Yep. It's still broken in the same way.

harrysarson:
Interesting, thank you @boxed!

PR #442, opened 2020-09-27
Replace reading .elmi files with a faster and less hacky approach

lydell:
- Runs tests 0.5 seconds faster generally, the same speed or possibly a little bit slower in the worst (uncommon) case.
- CI on this repo runs seems to run twice as fast!
- Gets rid of the `elmi-to-json` binary. No longer need to rely on undocumented, internal binary formats.
- On the other hand, we now have a custom Elm parser to maintain.
- And rely on the internal JS output of the compiler. At least it feels like this depends on fewer implementation details than the .elmi approach.
- More details: https://github.com/rtfeldman/node-test-runner/pull/442#issuecomment-713526271
- Status: Stable enough for general use.

More about speed:

- Both https://github.com/elm-in-elm/compiler/ and https://github.com/jfmengels/elm-review tests run 0.5 seconds faster on my machine.
- @sporto reported that it worked fine and fast on his project as well, where `elm-test 'src/**/*Test.elm'` is used rather than `tests/`. https://discourse.elm-lang.org/t/do-you-put-tests-outside-the-tests-folder/6386/11
- @bdukes reported it works fine and fast on Windows. https://discourse.elm-lang.org/t/do-you-put-tests-outside-the-tests-folder/6386/24
- If you run `elm-test src/`, this PR results in a much bigger JS file (since we don’t know until runtime what is and isn’t a `Test`). In this uncommon case, the wins from faster compilation time is eaten by slower execution time. Can be improved by improving parallelization in the future.

Post merge follow up:

- Make an issue about moving default seed generation to a much higher level of test generation. (~~ edited by @harrysarson)

<details>
<summary>Original PR description (outdated)</summary>

This is a proof of concept for @harrysarson’s idea to determine what exposed values are tests at runtime (using JavaScript) instead of at compile time (using elmi-to-json).

Pros:

- No longer need to rely on undocumented, internal binary formats.
- Should be a fix for https://github.com/rtfeldman/node-test-runner/issues/385 (requiring elmi-to-json is slow on Windows)
- Might be faster than compiling an extra time and then running elmi-to-json.

Cons:

- Need to rely on the internal output JS. At least it feels like this depends on fewer implementation details than the .elmi approach.
- Need an Elm code parser.
- Might be slower than compiling an extra time and then running elmi-to-json.

TODO:

- Replace the POC parser for finding which values a given .elm file exposes.
- More tests.
- Benchmark / Test in some real-world projects to see if it works out and if there’s any speed difference. 

Notes:

- elmi-to-json is still used to find the exact dependencies of package projects. This should be replaceable with `elm-json solve`. https://github.com/rtfeldman/node-test-runner/pull/356 adds elm-json, but not yet for finding exact dependencies of package projects.

</details>


.eslintrc.json
```diff
@@ -5,15 +5,17 @@
         "node": true
     },
     "parserOptions": {
-        "ecmaVersion": 6
+        "ecmaVersion": 2018
```
lydell, line 8: Needed for Unicode property escapes in regexps: `/\p{L}/u`.

This requires Node.js 10 or later – which is why I’ve removed Node.js 8 from supported versions in other files. Node.js 8 is EOL anyway so I think it should be fine.

elm/src/Test/Runner/Node.elm
```diff
@@ -309,11 +309,49 @@ init { processes, globs, paths, fuzzRuns, initialSeed, report, runners } _ =
...
+                Test.todo
+                    (if List.isEmpty paths then
+                        "I couldn't find any exposed values of type Test in any *.elm files in the tests/ directory of your project's root directory.\n\nTo generate some initial tests to get things going, run elm-test init"
+
+                     else
+                        "I couldn't find any exposed values of type Test in files matching:\n"
+                            ++ String.join "\n" paths
+                            ++ "\n\nMaybe try running elm-test with no arguments?"
+                    )
+
```
lydell, line 351: This check has moved from JS to Elm. I used `Test.todo` to get going with this experiment, but it feels a bit like a hack and I’m not sure if it gives the best output:

```
elm-test 0.19.1-revision4
-------------------------

Running 1 test. To reproduce these results, run: elm-test --fuzz 100 --seed 343653604167020 src


TEST RUN INCOMPLETE because there is 1 TODO remaining

Duration: 170 ms
Passed:   0
Failed:   0
Todo:     1
◦ TODO: I couldn't find any exposed values of type Test in files matching:
/Users/lydell/forks/node-test-runner/example-application/src/Something.elm

Maybe try running elm-test with no arguments?
```

example-package-no-core/elm.json
```diff
@@ -1,8 +1,12 @@
 {
   "type": "package",
-  "name": "example ",
+  "name": "elm-explorations/node-test-runner-example-package",
   "summary": "SSCCE on replicating issue 273",
+  "license": "BSD-3-Clause",
   "version": "4.0.1",
+  "exposed-modules": [
+    "Something"
+  ],
```
lydell, line 9: This package elm.json was broken so it didn’t test what it was supposed to.

example-package-no-tests/elm.json
```diff
@@ -5,7 +5,7 @@
     "license": "BSD-3-Clause",
     "version": "1.0.0",
     "exposed-modules": [
-        "Example"
+        "Something"
```
lydell, line 8: This example is supposed to fail because of no tests, not because of wrong `exposed-modules`.

lib/Generate.js
```diff
@@ -34,6 +34,22 @@ function prepareCompiledJsFile(pipeFilename /*: string */, dest /*: string */) {
...
+  return (
+    'var __elmTestSymbol = Symbol("elmTestSymbol");\n' +
+    content
+      .replace(testVariantDefinition, '$&, __elmTestSymbol: __elmTestSymbol')
+      .replace(
+        checkDefinition,
+        '$1return value && value.__elmTestSymbol === __elmTestSymbol ? $2Just(value) : $2Nothing;$3'
+      )
+  );
+}
```
lydell, line 51: Here are the JS hacks relying on the JS output of Elm. Also note that if the `Test` type every gets more variants, we need to add them to the first regex.

lib/Generate.js
```diff
@@ -414,4 +339,110 @@ function generateMainModule(
...
+  seed /*: number */,
+  report /*: string */,
+  testFileGlobs /*: Array<string> */,
+  testFilePaths /*: Array<string> */,
+  processes /*: number */
+) /*: string */ {
+  // TODO: CLI args should be parsed, validated and defaulted properly in elm-test.js.
+  const finalSeed = isNaN(seed)
+    ? Math.floor(Math.random() * 407199254740991) + 1000
+    : seed;
```
lydell, line 414: It’s super weird to create the default seed deep into the code generation routine. But this is out of scope for this PR.

lib/Parser.js
```diff
@@ -0,0 +1,81 @@
...
+    //       exposed.push('something');
+    //     } else {
+    //       stop();
+    //     }
+    //   });
+    //   reader.on('close', () => {
+    //     resolve(exposed);
+    //   });
+
+    // TODOx: POC parser for now. Does not support comments and multiline strings and is too liberal.
```
lydell, line 54: I haven’t figured out what the best way to parse this is yet. It would be really easy if it wasn’t for:

- multiline (possibly nested) comments
- multiline strings
- allowing newlines between tokens

lib/Runner.js
```diff
@@ -71,133 +47,87 @@ function getIndirectDeps(projectRootDir /*: string */) /*: Promise<Object> */ {
...
-function filterMap(array, f) {
-  return array.map(f).filter(Boolean);
+        default:
+          // This shouldn’t be possible for package projects.
+          return Promise.reject(
+            new Error(
+              multipleSourceDirectoriesError(filePath, matchingSourceDirs)
+            )
+          );
+      }
```
lydell, line 80: I wrote this out of necessity (previously we got module names from elmi-to-json. Now we need to figure them out ourselves), but it has a nice bonus effect: If a user runs `elm-test custom-test-dir/` and `custom-test-dir/` isn’t in `"source-directories"` they will now get a much better error message.

lib/Runner.js
```diff
@@ -71,133 +47,87 @@ function getIndirectDeps(projectRootDir /*: string */) /*: Promise<Object> */ {
...
-        } else {
-          return Promise.reject(
-            filePath +
-              ' has an invalid module declaration. Check the first line of the file and make sure it has a valid module declaration there!'
-          );
-        }
-      });
-    })
-  );
+Edit "source-directories" in your elm.json and try to make it so no source directory contains another source directory!
```
lydell, line 126: This message is inspired by Elm’s error message when you have two similar source directories.

lib/elm-test.js
```diff
@@ -92,26 +91,23 @@ function resolveFilePath(filename) {
...
   try {
     pathToElmBinary = path.resolve(which.sync(args.compiler));
-
-    elmVersion = spawn
-      .sync(pathToElmBinary, ['--version'], { encoding: 'utf-8' })
-      .stdout.trim();
   } catch (error) {
-    throw new Error('The elm executable passed to --compiler must exist.');
+    console.error('The elm executable passed to --compiler must exist.');
+    process.exit(1);
```
lydell, line 110: These errors were uncaught, which resulted in ugly console output (with full stack trace and all). Now they’re handled the same way as when we validate `args.reporter` somewhere else in this file.

lib/elm-test.js
```diff
@@ -332,17 +328,22 @@ if (args._[0] === 'make') {
...
+            pathToElmBinary,
+            testFilePaths,
+            packageIndirectDeps,
+            hasBeenGivenCustomGlobs
+          );
+        });
+      })
+      .catch((err) => {
+        console.error(err.message);
+        process.exit(1);
```
lydell, line 345: This used to fail with ”UnhandledPromiseRejectionWarning”. In a couple of other places as well.

tests/ci.js
```diff
@@ -10,70 +10,60 @@ const elmTestVersion = packageInfo.version;
...
+  shell.echo('### stderr');
+  shell.echo(result.stderr || '(no stderr)');
+  shell.echo();
+  shell.echo('### message');
+  shell.echo(message);
+  shell.echo(`Exit code: ${result.code}. See stdout and stderr above.`);
+  shell.echo(`This message comes from tests/${filename}`);
+  shell.echo();
+  shell.exit(1);
+}
```
lydell, line 65: This helped a lot when trying to get all tests to pass again.

.eslintrc.json
harrysarson: My policy (about which I could be persuaded) is that dropping node 8 needs to wait till elm 0.19.2 when it comes. Out of interest: why do we need to regex for unicode?

tests/ci.js
```diff
@@ -142,39 +132,39 @@ assertTestFailure();
...
 
 shell.cd('../');
 
-shell.echo('### Testing elm-test on example-package-no-core');
+shell.echo('\n### Testing elm-test on example-package-no-core');
 
 shell.cd('example-package-no-core');
 
-assertTestSuccess();
+assertTestErrored();
```
lydell, line 147: I don’t know why this one was marked as success before, and why it passed before. Running `elm make` in example-package-no-core yields:

```
-- MISSING DEPENDENCY ------------------------------------------------- elm.json

I need to see an "elm/core" dependency your elm.json file. The default imports
of `List` and `Maybe` do not work without it.

If you modified your elm.json by hand, try to change it back! And if you are
having trouble getting back to a working elm.json, it may be easier to find a
working package and start fresh with their elm.json file.
```

So it makes sense that it should fail IMO.

.eslintrc.json
lydell: To be able to match variable names. I’ll show where in the code this is.

lib/Parser.js
```diff
@@ -0,0 +1,81 @@
+const fs = require('fs-extra');
+
+// First char: https://github.com/elm/compiler/blob/2860c2e5306cb7093ba28ac7624e8f9eb8cbc867/compiler/src/Parse/Variable.hs#L296-L300
+// Rest: https://github.com/elm/compiler/blob/2860c2e5306cb7093ba28ac7624e8f9eb8cbc867/compiler/src/Parse/Variable.hs#L328-L335
+// https://hackage.haskell.org/package/base-4.14.0.0/docs/Data-Char.html#v:isLetter
+const nonCapitalizedNamePart = /\p{Ll}[_\d\p{L}]*/u;
```
lydell, line 6: @harrysarson Here’s how to match an Elm variable name.

Now that I think of it, this also matches reserved names such as `module`, `port` and `if`. I should detect that so the user doesn’t get a really bad error message if the do `port = 3000` by mistake.

lydell: I tried it out.

Main.elm:

```elm
module Main exposing (..)

import Bad

x = Bad.port
```

Bad.elm:

```
module Bad exposing (..)

port = 5
```

Result:

```
❯ elm make src/Main.elm
-- PORT PROBLEM ---------------------------------------------------- src/Bad.elm

I just saw the start of a `port` declaration, but then I got stuck here:

3| port = 5
        ^
I was expecting to see a name like send or receive next. Something that starts
with a lower-case letter.

Note: Here are some example `port` declarations for reference:

    port send : String -> Cmd msg
    port receive : (String -> msg) -> Sub msg

The first line defines a `send` port so you can send strings out to JavaScript.
Maybe you send them on a WebSocket or put them into IndexedDB. The second line
defines a `receive` port so you can receive strings from JavaScript. Maybe you
get receive messages when new WebSocket messages come in or when the IndexedDB
is changed for some external reason.
```

So it might not be necessary to detect reserved words – Elm seems to look for errors in imports first so the good errors come first. But that’s a bit of an implementation detail to rely on.

harrysarson:
Thanks @lydell, this is really exciting! A couple of thoughts up front:

Realistically we will merge the elm-json PR before this (#356) so you will (after rebasing) be able to drop the dependency here.

This PR touches quite a bit of the test runners nasty internal machinery so we will have to be careful and also strike the right balance between not changing too much too reduce the risk of breakage and not changing enough that the existing troublesome parts of the test runner cause issues! (For example, we have had complaints that process.aborting causes corrupted output because the stream doesn't get time to flush, I see this PR adds more process.abort -- not a bad thing -- so we need to be careful that we don't make the problem worse.)

I will endeavour to review soon! Tagging @mpizenberg as I think this may be interesting to you :)

.eslintrc.json
lydell: Either way, if supporting Node.js 8 is desired we could use a worse regex on Node.js 8, which means Node.js 8 at least gets partial support.

lydell:
What do you mean by `process.abort`?

mpizenberg:
Interesting indeed!

harrysarson:
I mean process.exit! Will edit :)

lydell:
Ah, I see!

2 of the added `process.exit` are for validating the `--compiler` flag. That happens very early, before any streams or external processes have been started.

The remaining 2 are in places where we used to exit with “UnhandledPromiseRejection”. It’s unclear what happened before in those cases. But if anything I’d guess the new behavior is the same or at least not worse. Better handling with cleanup and stuff could be added in another PR.

harrysarson:
That makes sense! Consider the process.exit concern resolved :+1:

lib/Parser.js
harrysarson: Is this too evil https://mothereff.in/regexpu#input=const+nonCapitalizedNamePart+%3D+%2F%5Cp%7BLl%7D%5B_%5Cd%5Cp%7BL%7D%5D*%2Fu%3B&dotAllFlag=1&unicodePropertyEscape=1?

lydell: Yeah, we _could_ do that if we want to support Node.js 8. We could also use `[a-z][_\da-zA-Z]*` for Node.js 8 and call it a day.

harrysarson:
Bad news :worried: 

image

With this PR, we only find a fraction of the number of tests as without it. (I am testing on my copy of https://github.com/elm-in-elm/compiler/). Will investigate futher!

example-package-no-core/elm.json
harrysarson: Question is what is it supposed to test? 

This is the issue I fixed and added this as a regression test for: https://github.com/rtfeldman/node-test-runner/issues/273
@Maxim-Filimonov put together a SSCCE [here](https://github.com/Maxim-Filimonov/node-test-runner-273-sscce) but looking at that SSCCE the elm.json file has no dependancy on elm/core so it should fail! @Maxim-Filimonov I realise it was over a year ago that you made the SSCCE but can you remember why you expected a package with no dependancy on elm/core to pass testing?

lib/Generate.js
harrysarson: Could we consider instead changing the names of the custom type constructors (the definition is [here](https://github.com/elm-explorations/test/blob/b639def0fdb3437cd076d4ed840cb2451c553f7c/src/Test/Internal.elm#L10)) like so:

```elm


type Test
    = ElmTestVariant__UnitTest (() -> List Expectation)
    | ElmTestVariant__FuzzTest (Random.Seed -> Int -> List Expectation)
    | ElmTestVariant__Labeled String Test
    | ElmTestVariant__Skipped Test
    | ElmTestVariant__Only Test
    | ElmTestVariant__Batch (List Test)
```

and then we could use the regex to check for the `ElmTestVariant__` bit. That way it should be obvious to future contributors to the test library that they need to add an `ElmTestVariant__` and the test runner would immediately pick it up?

lib/Generate.js
harrysarson: I have edited the PR description with a follow up task to make an issue to move it :+1:

example-package-no-core/elm.json
harrysarson: I am thinking we can delete this example.

example-package-no-tests/elm.json
harrysarson: Thanks for catching!

lib/Runner.js
harrysarson: Nice!

lib/elm-test.js
harrysarson: Thanks!

tests/ci.js
harrysarson: This will make my life my easier too going forward!

harrysarson:
For reference this is my original sketch of this idea: https://gist.github.com/harrysarson/08acee2e8d8ffb0185b4b1020c08d756

lib/Parser.js
harrysarson: I think the "detect tests based on js" approach can only work if it does not require maintaining an elm source code parser within elm-test. Additionally, I do not think a regex based approach can ever be robust enough here. I wonder if we can pull in an external dependency here. What do you think @lydell? If I had to guess where the missing tests in https://github.com/rtfeldman/node-test-runner/pull/442#issuecomment-702983427 have gone I would say that these regexes miss them,

lib/Generate.js
lydell: That sounds like a good idea! I can make a PR to elm-explorations/test as well. We can support both the old and new names in the regex so the runner works with both old and new versions of the test framework.

lib/Parser.js
lydell: - A regex approach is never going to be robust enough, that’s correct. I just did it with regex so I could get something working to validate the overall approach.
- I’m already working a a custom parser, that parses just enough for elm-test’s needs. It’s fun to write so I plan on finishing it even if we don’t end up using it in the end. I’ll let you know when I have more to show.
- The only Elm parser on npm I know of is tree-sitter-elm. I played around with it quickly before doing the regexes. Unfortunately it took almost half a minute to install (postinstall-compile) on my macBook. And my _personal_ motivation for this PR is to reduce install time! 😢

lib/Generate.js
harrysarson: Brilliant!

lib/Parser.js
harrysarson: Okay! Good points, I will wait and see what you create and reserve any judgement until then!

lydell:
Good news!

I got completely into the zone and just banged out a parser. I hope I still understand the code once I get out of the zone again.

Surprisingly, it worked on the elm-in-elm compiler tests on the first try! Even though I haven’t even written tests for the parser yet! 🤯 


```
~/forks/compiler master
❯ npx elm-test

elm-test 0.19.1
---------------

Running 402 tests. To reproduce these results, run: elm-test --fuzz 100 --seed 389628710478272 /Users/lydell/forks/compiler/tests/DesugarTest.elm /Users/lydell/forks/compiler/tests/EmitJsonTest.elm /Users/lydell/forks/compiler/tests/EmitTest.elm /Users/lydell/forks/compiler/tests/InferTypesFuzz.elm /Users/lydell/forks/compiler/tests/InferTypesTest.elm /Users/lydell/forks/compiler/tests/LocatedTest.elm /Users/lydell/forks/compiler/tests/OptimizeTest.elm /Users/lydell/forks/compiler/tests/ParserTest.elm /Users/lydell/forks/compiler/tests/Shrink/Extra.elm /Users/lydell/forks/compiler/tests/TestHelpers.elm

↓ ParserTest
↓ Stage.Parse.Parser.valueDeclaration
✗ user defined type with argument - newline only

    x : Foo
    ()
    x = ()
    ===> should have failed but parsed into ==>
    Ok
        ("x",Value { expression = Unit, typeAnnotation = Just { type_ = UserDefinedType { args = [Unit], name = "Foo", qualifiedness = PossiblyQualified Nothing }, varName = "x" } })



TEST RUN FAILED

Duration: 1080 ms
Passed:   401
Failed:   1


~/forks/compiler master
❯ ../node-test-runner/bin/elm-test

elm-test 0.19.1-revision4
-------------------------

Running 402 tests. To reproduce these results, run: elm-test --fuzz 100 --seed 23432219873072

↓ ParserTest
↓ Stage.Parse.Parser.valueDeclaration
✗ user defined type with argument - newline only

    x : Foo
    ()
    x = ()
    ===> should have failed but parsed into ==>
    Ok
        ("x",Value { expression = Unit, typeAnnotation = Just { type_ = UserDefinedType { args = [Unit], name = "Foo", qualifiedness = PossiblyQualified Nothing }, varName = "x" } })



TEST RUN FAILED

Duration: 1078 ms
Passed:   401
Failed:   1

```

Same amount of tests run, and in same time! Made me wonder if I’m really running my branch, but when I introduced a bug in the parser on purpose it did get triggered so it seems to be running.

The parser is currently ~150 lines of stateful mutating code, and ~600 lines of comments, constants and pure functions.

I’m going to write a couple of parser tests next, to validate its approach. I can write much more tests later if we decide to use this parser. Until then I’ll spend my time on benchmarking and running tests in other projects.

lib/Generate.js
lydell: https://github.com/elm-explorations/test/pull/153

lydell:
Exciting update: I’ve written before that this PR takes the same amount of time to execute tests as the master branch. Turns out I was mistaken!

Here’s an example test report:

```
TEST RUN PASSED

Duration: 2247 ms
Passed:   618
Failed:   0
```

Turns out the “Duration” there is only counting the time it takes to execute the compiled JS for the tests – not the time it takes to generate that compiled JS!

So that explains the extremely similar durations between this PR and master. Both execute the same thing so of course it takes the same time. (Actually, the compiled code in this PR also needs to filter out values that aren’t `Test`s, but it looks like that takes no time at all.)

In both elm-in-elm/compiler and jfmengels/elm-review I’m seeing **this PR take roughly 1 second less** for the whole thing!

Even when running `elm-test` on the unmodified output of `elm-test init` is a little bit faster with this PR!

What I’ve got left to time is on big projects that don’t use a `tests` folder and instead put the tests among the source code. That means many more files to parse – I’m curious to see how it performs.

lydell:
I added one single `1 + 1 == 2` test in a random file in `src/` at our 60K SLOC code base at work and ran `elm-test src`. These are the results:

| branch | full duration | execution duration |
|-|-|-|
| master | 5.3 s | 150 ms |
| this PR | 3.3 s | 470 ms |

As you can see, with this amount of code the filtering of `Test` values now starts to show. But overall it was still 2 seconds faster!

A downside I guess is that the execution duration now includes executing even more code than before that aren’t your own test code. So at this scale the number starts losing its value maybe.

Now that I think of it … maybe we need to rethink the duration timing overall. Maybe it should time the whole thing, not just the execution? I remember seeing `Duration: 150 ms` in `elm-test` output and thinking to myself “Hmm, it felt like seconds?”. But even if we start the timer as soon as we can, we still miss the infamous 200ms Node.js startup time. Is it better to let people run `time elm-test` if they want duration? 🤔

.eslintrc.json
lydell: Linking to https://github.com/rtfeldman/node-test-runner/issues/416 for reference.

Note that Node.js 8 became EOL 10 months ago – 2019-12-31.

Supporting EOL Node.js versions sounds like a bad idea to me. Elm 0.19.2 might be years away. By then, at least Node.js 10 will be EOL as well (2021-04-30). Supported Elm version and supported Node.js version has no connection in my mind.

- Upgrading Node.js is super easy these days anyway, IMO.
- And if someone doesn’t want to upgrade and is fine running an EOL Node.js, then they should be fine running an older elm-test as well. It’s not like the old version would stop working for them.

lib/Generate.js
lydell: Possibly related: https://github.com/rtfeldman/node-test-runner/issues/248

lib/Parser.js
```diff
@@ -0,0 +1,788 @@
...
+      if (token.tag === 'NewChunk') {
+        return 'NewChunk';
+      }
+      return expected('a new chunk', token);
+
+    case 'NewChunk':
+      if (token.tag === 'Other') {
+        switch (token.value) {
+          case 'port':
+          case 'effect':
```
lydell, line 610: So this tries to support `effect module` but I just learned that those look like this:

```elm
effect module ModuleName where { subscription = MySub } exposing (myVal)
```

So to support that for real there’s additional syntax to parse (`where { subscription = MySub }`). Only blessed modules can use `effect module`, right? Do we know if any of them uses elm-test and needs elm-test to be able to parse the `effect module` files to find all tests?

.eslintrc.json
harrysarson: What we should have done was to drop node 8 support when we released 0.19.1. We should definitely make sure that we drop support for enough node versions when we drop 0.19.2 so that we avoid this problem in the future.

> -   Upgrading Node.js is super easy these days anyway, IMO.

True

> -   And if someone doesn’t want to upgrade and is fine running an EOL Node.js, then they should be fine running an older elm-test as well. It’s not like the old version would stop working for them.

Also true.

lib/Parser.js
lydell: The conclusion is:

Only packages can have `effect module` (and on top of that, only `elm/` and `elm-explorations/` packages), inside `src/`. Packages can only have tests inside `src/` if they depend directly on elm-explorations/test, which isn’t practical. So there’s no need to support effect modules. Instead we exit elm-test if we encounter one.

I’m going to require package tests to be in `tests/` as a separate commit.

example-package-no-core/elm.json
lydell: Removed in https://github.com/rtfeldman/node-test-runner/pull/448

example-package-no-tests/elm.json
lydell: Merged in https://github.com/rtfeldman/node-test-runner/pull/448

lib/elm-test.js
lydell: Merged in https://github.com/rtfeldman/node-test-runner/pull/448

tests/ci.js
lydell: Merged in https://github.com/rtfeldman/node-test-runner/pull/448

tests/ci.js
lydell: Test removed in https://github.com/rtfeldman/node-test-runner/pull/448

.eslintrc.json
lydell: Copying in stuff from comments in another thread (so I can minimize the other one).

These are our choices:

1. Drop Node.js 8 support.
2. Fallback to this in Node.js 8: https://mothereff.in/regexpu#input=const+nonCapitalizedNamePart+%3D+%2F%5Cp%7BLl%7D%5B_%5Cd%5Cp%7BL%7D%5D*%2Fu%3B&dotAllFlag=1&unicodePropertyEscape=1
3. Use `[a-z][_\da-zA-Z]*` in Node.js 8.

lib/Generate.js
lydell: Also related: https://github.com/rtfeldman/node-test-runner/issues/211

lib/Solve.js
```diff
@@ -0,0 +1,72 @@
...
+  pathToElmJson /*: string */,
+  projectElmJson /*: any */
+) /*: { direct: { [string]: string }, indirect: { [string]: string } } */ {
+  // If possible, use a simple and fast merge of all dependencies in elm.json,
+  // otherwise use elm-json. At the time of writing, this could save 1 second in
+  // an application project.
+  const simpleMerge = trySimpleMerge(projectElmJson);
+  if (simpleMerge !== undefined) {
+    return simpleMerge;
+  }
```
lydell, line 17: @zwilias As our elm dependencies expert, would you mind checking if `trySimpleMerge` is sound, or if it can cause errors? If it works this is a pretty great speedup.

mpizenberg: You should be able to merge them if they are unique indeed, but under the assumption that indirect dependencies and indirect test dependencies are correct. This is fairly often the case for project indirect dependencies since they are checked by elm. But for the indirect test dependencies, it is a bit more difficult to trust (especially for people not using `elm-json` to install their test dependencies).

lydell: I don’t think it matters much if dependencies are invalid because the error you get is OK:

```
~/forks/node-test-runner/example-application js *$
❯ ../bin/elm-test
-- INCOMPATIBLE DEPENDENCIES ------------------------------------------ elm.json

The dependencies in your elm.json are not compatible.

Did you change them by hand? Try to change it back! It is much more reliable to
add dependencies with elm install or the dependency management tool in
elm reactor.

Please ask for help on the community forums if you try those paths and are still
having problems!
```

It’s basically the same error as if you try to compile your project with `elm make`.

lydell: This is the error you get (from elm-json) if disabling my shortcut:

```
~/forks/node-test-runner/example-application js *$
❯ ../bin/elm-test

-- NO VALID PACKAGE VERSION ----------------------------------------------------

Because elm-explorations/test 1.2.0 depends on elm/virtual-dom 1.0.0 <= v <
2.0.0 and elm/virtual-dom 1.0.0 <= v < 2.0.0 is unavailable, no valid set of
package versions could be found.

And because this project depends on elm-explorations/test 1.2.0, no valid set of
package versions could be found.
```

Not sure it’s much better?

mpizenberg: This is because elm-json is used in offline mode right? does it do the same in online mode?

lydell: As far as I know, elm-json is only in offline mode if you use the `--offline` flag. That’s not present in this file. And I’m connected to the Internet. But maybe my example is bad. I made the elm.json invalid by saying I depend on elm/virtual-dom 2.0.0 which doesn’t exist.

mpizenberg: I see, maybe try removing an indirect dependency to see if there is a difference

lydell: I removed elm/virtual-dom from indirect. Elm fails with the usual message, while `elm-json solve` adds it back, making the tests run. But if you’d try to compile your code you’ll of course get Elm’s error. So maybe it’s weird that the tests work, when compiling the project doesn’t? Or rather – I don’t think it matters that the fast version doesn’t make it since your project is kind of invalid anyway.

mpizenberg: Oh yeah sorry I meant remove something from indirect test dependencies. Since most often (before elm-json) people would just add some library to their test dependencies if they need them for tests, but would not search for the indirect ones. In that case, elm-test used to work, and works with elm-json. But it could make sense to fail if the dependencies are incorrect.

lydell: Ok, I used `elm-test install truqu/md5` which correctly added that to direct test-dependencies, as well as `zwilias/elm-utf-tools` to _indirect_ test-dependencies. I added a `MD5.hex` call to the tests just to make sure everything worked.

Then i removed `zwilias/elm-utf-tools` from elm.json. Same thing as before – `elm-json solve` adds it back, so tests pass, while `elm make` or my fast version fail with the “were edited by hand” error.

lib/Parser.js
```diff
@@ -0,0 +1,809 @@
...
+    let newChunk = true;
+
+    let lastLowerName = '';
+
+    let index = 0;
+
+    const readable = createReadStream(filePath, {
+      encoding: 'utf8',
+      highWaterMark: 4096,
+    });
```
lydell, line 54: This number is completely arbitrary.

According to the [createReadStream docs](https://nodejs.org/api/fs.html#fs_fs_createreadstream_path_options) the default is 64 kB. And the default for the readable stream base class is 16 kB. So 4 kB is way less.

At work 4.0 kB is by far the most common size of all our Elm files (the block size of the file system is 4 kB on macOS):

```
 158 4.0K
  49 8.0K
  20  12K
  14  16K
  13  20K
   6  24K
   5  28K
   2  36K
   4  44K
   1  48K
   1  52K
   2  56K
   1  60K
   1  68K
   1  76K
   1  80K
   1  88K
   1 112K
   1 172K
```

So reading 4 kB at a time reads the entire file most of the time. Which maybe either is fine, or excessive.

I have tried setting it to `1` (which should read 1 char at at time). That seems to be slightly slower. But basically any other number seems to be faster and perform the same.

So maybe it’s all about memory usage. If 4 kB is read at a time per file, you’d need 256 files to use 1 MB of memory. That’s very little, right? At work we have 282 Elm files currently. But way fewer in `tests/`.

At work our average module declaration is 120 characters. Mean 80. Smallest 27, largest 1214. But in `tests/` they only range between 32 and 49.

So it might make sense to have a highWaterMark around that. Unless people usually have `exposing (..)` in their tests. Then it’s maybe more sensible to aim for the average test name length or something. But my guess would be that people expose either everything (`(..)`) or just one or a couple of values per file.

`find-elm-dependencies` (which is used (indirectly) by elm-webpack-loader) uses 60 chars at a time by reading 480 bytes per chunk: https://github.com/NoRedInk/find-elm-dependencies/blob/c355c864e1bd334998d27d750c9fd38558730f39/src/dependencies.js#L8-L9

In conclusion:

- It does not generally affect performance? Maybe when not having an SSD?
- Is it worth chasing the ideal memory usage?
- Can we measure the idea number?

harrysarson: So we use this stream to read from an elm file, then we parse the incoming data and produce definitions.

I would not be surprised if we are IO bound here. Certainly I would be surprised if we ever had to ask the OS to stop giving us data because we are processing it too slowly. I think (I may be wrong) that this means that we should never reach the high watermark provided it isn't set to something really small.

lydell: > So we use this stream to read from an elm file, then we parse the incoming data and produce definitions?

Correct! And unless `exposing (..)` is used, we stop the stream once we’ve parsed the module declaration.

harrysarson: I think go with the default value then and trust that it will not matter in practice

lydell: Switched to default.

elm/src/Test/Runner/Node.elm
```diff
@@ -309,11 +309,52 @@ init { processes, globs, paths, fuzzRuns, initialSeed, report, runners } _ =
...
 
+{-| The implementation of this function will be replaced in the generated JS
+with a version that returns `Just value` if `value` is a `Test`.
+
+If you rename this function you also need to update the regex that looks for it.
+
+-}
+check : a -> Maybe Test
+check value =
+    Nothing
```
harrysarson, line 320:
```suggestion
check =
    \value -> check value
```

I think a crash is better than silently failing if somehow this function is not replaced in the js.

lydell: Wouldn’t `Debug.todo "Informative message"` be better?

harrysarson: I forget that we are allowed to use `Debug.todo`. Yes it would!

lydell: This makes the `.replace()` a bit more difficult but I’ll see what I can do.

lydell: Now using Debug.todo

elm/src/Test/Runner/Node.elm
```diff
@@ -309,11 +309,52 @@ init { processes, globs, paths, fuzzRuns, initialSeed, report, runners } _ =
...
 
+{-| The implementation of this function will be replaced in the generated JS
+with a version that returns `Just value` if `value` is a `Test`, otherwise `Nothing`.
+
+If you rename or change this function you also need to update the regex that looks for it.
+
+-}
+check : a -> Maybe Test
+check value =
+    Debug.todo "The regex for replacing this Debug.todo with some real code must have failed since you see this message!\n\nPlease report this bug: https://github.com/rtfeldman/node-test-runner/issues/new\n"
```
harrysarson, line 320: You are probably fed up with my never ending suggestions about this function so please feel free to ignore me. This might be the best of both worlds though (crashes if our regex fails and should allow a simply regex):

```suggestion
check : a -> Maybe Test
check =
    checkHelperReplaceMe___
    
 
checkHelperReplaceMe___ : a -> b
checkHelperReplaceMe___ _=
    Debug.todo "The regex for replacing this Debug.todo with some real code must have failed since you see this message!\n\nPlease report this bug: https://github.com/rtfeldman/node-test-runner/issues/new\n"
```

lydell: Great idea! :+1:

drathier:
Could someone write up a description of why the elm test variants should be prefixed? What does the regex hack do? How does it work? What's the goal of the change in isolation, and of this whole PR? How much is 0.5s speed increase in relative speed-up? Is it worth the regex hack?

I'm mostly working in the elm-explorations/test repo, so I don't know any details about this other than this PR. I want a description we can link to so we know why the elm code is the way it is.

lydell:
@drathier Here are some comments in the code:

https://github.com/elm-explorations/test/pull/153/files#diff-8772bcfce271f7a0c17ff1b886a61573ba5707f5d75e1ad15646f4ac71e2e22fR10-R13

https://github.com/rtfeldman/node-test-runner/pull/442/files#diff-a6b9b7d772c648c3de40bac1b5132fea11f74f861f6180c43e9457650e403e07R312-R320

https://github.com/rtfeldman/node-test-runner/pull/442/files#diff-0753c3cb173ce267ac31a39601e397e649dd117794b99db5064b83f521405537R40-R58

This PR is trading one hack for another, when it comes to finding exposed `Test` values.

The old hack is copying a bunch of the Elm compiler source code into a separate project – elmi-to-json – and slimming it down to make it only read the internal .elmi files that Elm produces. We invoke the Elm compiler one extra time to produce the .elmi files as a side effect and then read them using `elmi-to-json`.

The new hack is modifying the JS output from Elm a little. We tag all `Test` constructors with a symbol, and modify a dummy function to look for that symbol to know if its parameter is a `Test` or not. So we’re checking what is a test and what isn’t at runtime rather than at compile time.

Pros:

- Better performance. I’ll have to get back to you with relative speed-up.
- CI on this repo runs seems to run twice as fast!
- No need to download a separate binary. Which has been problematic on Windows: #385
- We depend on less implementation details. Instead of depending on the entire .elmi binary format we depend on a couple of details of Elm’s JS output.

Cons:

- We need to maintain a (partial) Elm parser.
- The ugly `ElmTestVariant__` prefix, I guess.

@harrysarson can probably give you more details – he’s been thinking about this for a long time. I mostly just implemented this gist of his idea :)

lydell:
This is what the regex is doing.

Tagging Test constructors:

```js
var $elm_explorations$test$Test$Internal$Batch = function (a) {
  return {$: 'Batch', a: a};
};
```

⬇️

```js
var $elm_explorations$test$Test$Internal$Batch = function (a) {
	return {$: 'Batch', __elmTestSymbol: __elmTestSymbol, a: a};
};
```

Repeat for every test variant. **If they all have a common prefix this is trivial and forwards-compatible.**

Replacing dummy function:

```js
var $author$project$Test$Runner$Node$check = $author$project$Test$Runner$Node$checkHelperReplaceMe___;
```

⬇️

```js
var $author$project$Test$Runner$Node$check = value => value && value.__elmTestSymbol === __elmTestSymbol ? $elm$core$Maybe$Just(value) : $elm$core$Maybe$Nothing;
```

lydell:
Some performance notes:

**elm-review**
master: 3.2 seconds
this pr: 2.6 seconds

**at work**
master: 1.5 seconds
this pr: 0.8 seconds

harrysarson:
My main reasons for wanting to migrate away from elmi-to-json:

1. Depending on the binary `.elmi` format means that every release of the elm compiler requires us to update the test runner. Depending on the generated javascript feels a bit less hacky. It is plain text and (based on my experience of elm 0.18 and 0.19) more stable than the `.elmi` format.
2. There is a tentative project to replace the test-runner with something written in rust. Whilst binwrap makes it quite easy for the test runner to depend on (and shell out to) a binary tool bundling a binary tool into a rust executable is not possible/very hard. I looked into statically link elmi-to-json into rust but the rust/haskell FFI was pretty gnarly and I eventually gave up.
3. This approach is a bit faster.
4. (A bit more tenuous.) Already out of the PR we have spun out and landed #448, #451, #453, #455, #457. Even if the "JS hack" approach does not ultimately go anywhere (but I am very confident it will) there is great value in pursuing this exploration. 

---

Aside: we should probably think of a better name for the "JS hack" approach for marketing purposes.

elm/src/Test/Runner/Node.elm
lydell: @harrysarson This is one of the few things left to do. Do we think the above message is good enough, or do you have ideas on how to handle this case better?

.eslintrc.json
harrysarson: Could you make a separate PR to drop node 8 support? (Add node 15 to the matrix at the same time if you like)

lydell: Sure!

lydell: PR: https://github.com/rtfeldman/node-test-runner/pull/461

elm/src/Test/Runner/Node.elm
lydell: I figured it out!

### Before I figured it out:

If the CLI arguments match no files:

```
❯ ../bin/elm-test 'src/**/*Test.elm'
No tests found for the file pattern "src/**/*Test.elm"

Maybe try running elm-test with no arguments?
```

If the CLI arguments do match files, but they contain no exposed Test values:

```
❯ ../bin/elm-test src

elm-test 0.19.1-revision4
-------------------------

Running 1 test. To reproduce these results, run: elm-test --fuzz 100 --seed 76422859129210 src


TEST RUN INCOMPLETE because there is 1 TODO remaining

Duration: 164 ms
Passed:   0
Failed:   0
Todo:     1
◦ TODO: I couldn't find any exposed values of type Test in files matching:
/Users/lydell/forks/node-test-runner/example-application/src/Something.elm

Maybe try running elm-test with no arguments?
```

### After

If the CLI arguments match no files:

```
❯ ../bin/elm-test 'src/**/*Test.elm'
No files found matching:

src/**/*Test.elm

Are the above patterns correct? Maybe try running elm-test with no arguments?
```

If the CLI arguments do match files, but they contain no exposed Test values:

```
❯ ../bin/elm-test src
No exposed values of type Test found in files matching:

src

Are the above patterns correct? Maybe try running elm-test with no arguments?

Are there tests in any of the matched files?
If not – add some!
If there are – are they exposed?
```

lydell:
Marking as “Ready for review” since I have no further ideas on things to improve (other than possibly adding a test for something – I’m all ears)! 🎉

elm/src/Test/Runner/Node.elm
```diff
@@ -309,33 +309,132 @@ init { processes, globs, paths, fuzzRuns, initialSeed, report, runners } _ =
...
+                , paths = []
+                , fuzzRuns = 0
+                , initialSeed = 0
+                }
+            , processes = 0
+            , nextTestToRun = 0
+            , results = []
+            , testReporter = createReporter report
+            , autoFail = Nothing
+            }
```
lydell, line 329: I’m not super happy about having to create an empty model that is never going to be used here, but it was the best I could come up with.

harrysarson (commented):
Some quick comments/questions. On the whole is very solid :)

I need to have a proper look through the parser still.

lib/Generate.js
```diff
@@ -309,4 +235,117 @@ function generateMainModule(
...
+      }
+  }
+}
+
+function makeElmString(string) {
+  return `"${string
+    .replace(/[\\"]/g, '\\$&')
+    .replace(/\n/g, '\\n')
+    .replace(/\r/g, '\\r')}"`;
+}
```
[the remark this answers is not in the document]
lydell: We still need to escape `\` and `"""` for triple quote string. Maybe even `\r` – not sure what Elm does with line endings exactly. So I don’t think the code gets easier from it.

lib/Generate.js
```diff
@@ -181,104 +196,17 @@ function generateMainModule(
...
-    const errorMessage =
-      testFilePaths.length > 0
-        ? 'I couldn\'t find any exposed values of type Test in files matching "' +
-          testFilePaths.toString() +
-          '"\n\nMaybe try running elm-test with no arguments?'
-        : "I couldn't find any exposed values of type Test in any *.elm files in the tests/ directory of your project's root directory.\n\nTo generate some initial tests to get things going, run elm-test init";
-
-    console.error(errorMessage);
-    process.exit(1);
-  }
```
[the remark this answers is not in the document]
lydell: Correct! I tried this the other day, and switching between 0 and 1+ exposed tests during watch works now!

lib/Generate.js
```diff
@@ -37,6 +37,26 @@ function prepareCompiledJsFile(
...
 
+// For older versions of elm-explorations/test we need to list every single
+// variant of the `Test` type. To avoid having to update this regex if a new
+// variant is added, newer versions of elm-explorations/test have prefixed all
+// variants with `ElmTestVariant__` so we can match just on that.
+const testVariantDefinition = /^var\s+\$elm_explorations\$test\$Test\$Internal\$(ElmTestVariant__\w+|UnitTest|FuzzTest|Labeled|Skipped|Only|Batch)\s*=\s*(?:\w+\(\s*)?function\s*\([\w, ]*\)\s*\{\s*return *\{\s*\$:\s*(['"])\1\2/gm;
+
+const checkDefinition = /^(var\s+\$author\$project\$Test\$Runner\$Node\$check)\s*=\s*\$author\$project\$Test\$Runner\$Node\$checkHelperReplaceMe___;?$/m;
+
+function hackCompiledElmJs(content) {
```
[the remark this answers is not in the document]
lydell: I agree `hackCompiledElmJs` is not a very good name, but the function does more than generating the check function. It also adds the __elmTestSymbol definition, tags all `Test` constructors with it and replaces the dummy check function with the “generated” one.

Maybe `implementKernelCheckFunction` or something? 🤔

lib/Generate.js
```diff
@@ -37,6 +37,26 @@ function prepareCompiledJsFile(
   });
 }
 
+// For older versions of elm-explorations/test we need to list every single
+// variant of the `Test` type. To avoid having to update this regex if a new
+// variant is added, newer versions of elm-explorations/test have prefixed all
+// variants with `ElmTestVariant__` so we can match just on that.
+const testVariantDefinition = /^var\s+\$elm_explorations\$test\$Test\$Internal\$(ElmTestVariant__\w+|UnitTest|FuzzTest|Labeled|Skipped|Only|Batch)\s*=\s*(?:\w+\(\s*)?function\s*\([\w, ]*\)\s*\{\s*return *\{\s*\$:\s*(['"])\1\2/gm;
```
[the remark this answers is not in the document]
lydell: Not that I know of.

lib/Generate.js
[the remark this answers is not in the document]
harrysarson: `addKernelTestChecking`? Naming things has a reputation for being the hardest part of software engineering for a reason.

lib/Generate.js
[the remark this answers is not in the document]
harrysarson: Fantastic!

lib/Generate.js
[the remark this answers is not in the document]
harrysarson: Chill. Makes sense!

Issue #354, opened 2019-04-04
Attribute spelling doesn't match spec - and causes jenkins parsing to fail


jhf:
When I try to make Jenkins read the Elm test results with JUnit, the parsing of the tests fails, with the
following error message:
```
INFO: [JUnit] - 1 test report file(s) were found with the pattern 'TestResultElm.xml' relative to '...' for the testing framework 'JUnit'.
WARNING: The file '.../TestResultElm.xml' is an invalid file.
WARNING: At line 1 of file:.../TestResultElm.xml:cvc-complex-type.3.2.2: Attribute 'failed' is not allowed to appear in element 'testsuite'.
WARNING: At line 1 of file:.../TestResultElm.xml:cvc-complex-type.4: Attribute 'failures' must appear on element 'testsuite'.
WARNING: The result file '.../TestResultElm.xml' for the metric 'JUnit' is not valid. The result file has been skipped.
```

Looking at the jUnit "spec", not that there is an official spec, but at least jUnit and ANT appears to agree, http://help.catchsoftware.com/display/ET/JUnit+Format shows that the attribute name should be changed to `failures`.

The line in question is here
https://github.com/rtfeldman/node-test-runner/blob/cdb59ac76cd69a183af3e4a1c9a5f65203fa047f/elm/src/Test/Reporter/JUnit.elm#L113

harrysarson:
PR welcome! Is there a way we can validate the JUnit we produce?

PR #445, opened 2020-10-01
Bugfix incorrect attribute name

jhf:
As per the spec, or the closest thing there is to one,
ref "https://help.catchsoftware.com/display/ET/JUnit+Format".

### DIFF
--- a/elm/src/Test/Reporter/JUnit.elm
+++ b/elm/src/Test/Reporter/JUnit.elm
@@ -110,7 +110,7 @@ reportSummary { testCount, duration, failed } autoFail =
 
                 -- Would be nice to have this provided from elm-package.json of tests
                 , ( "@tests", Encode.int testCount )
-                , ( "@failed", Encode.int failed )
+                , ( "@failures", Encode.int failed )
                 , ( "@errors", Encode.int 0 )
                 , ( "@time", Encode.float duration )
                 , ( "testcase", Encode.list identity extraFailures )


harrysarson:
Test failure is due to flakiness in CI do not worry about that.

This looks good to me. Just to continue the [conversation about validation](https://github.com/rtfeldman/node-test-runner/issues/354#issuecomment-702226054), what I am thinking is this:

1. Without this PR, when you run an elm-test test suite in jenkins, jenkins complains about the JUnit xml format.
2. With this PR Jenkins is (presumably) happy to run elm-test test suites!

Therefore, Jenkins must run some sort of program that parses the JUnit XML which can complain about the formatting. If this program was available and we could set it up in our CI (maybe just linux to keep it easy) running with some test output, we would have confidence that the XML elm-test produces is valid!

Saying that, if this all sounds like too much work I am happy to merge this as is.

harrysarson:
Re-opening to see if we can get CI green!

harrysarson:
@jhf do you have any more thoughts here?

jhf:
> Test failure is due to flakiness in CI do not worry about that.

Ok

> This looks good to me. Just to continue the [conversation about validation](https://github.com/rtfeldman/node-test-runner/issues/354#issuecomment-702226054), what I am thinking is this:
> 
> 1. Without this PR, when you run an elm-test test suite in jenkins, jenkins complains about the JUnit xml format.
> 2. With this PR Jenkins is (presumably) happy to run elm-test test suites!

Yes

> Therefore, Jenkins must run some sort of program that parses the JUnit XML which can complain about the formatting.

No, it tries to parse the xml, and fails if some required field is missing, do notice that Jenkins is not strict on all fields/attributes,
just those it cares about/expect, and it does not require a validated xml.

> If this program was available and we could set it up in our CI (maybe just linux to keep it easy) running with some test output, we would have confidence that the XML elm-test produces is valid!

Maybe, because the different tags and attributes depend on the outcomes, no failing tests, some failing tests, and on jenkins, we would need to produce a variety of test cases, and run the jenkins parser from the command line.
That should be possible, but it may require some Java programming to actually run the code independent of a complete Jenkins installation.

Furthermore, I would not have complete confidence in the output, even if Jenkins work, since Jenkins doesn't validate.

I think the best bet would be to take the XSD file I mentioned in [1] and find a standalone program that can validate the xml output, using the XSD file. An XSD file contains definitions for the validity of an xml.

Glancing over the XSD, I think it requires much more than Jenkins, in particular a timestamp of the time when running the test.

> Saying that, if this all sounds like too much work I am happy to merge this as is.

I think perfect is the enemy of good - in this case, having an xml that is at least not known to be wrong is better than now. Having a fully validated xml is better, from a purity standpoint, but will make difference in Jenkins parsing the file.

[1] https://github.com/rtfeldman/node-test-runner/issues/354#issuecomment-702226054

harrysarson:
Okay! You make good points and you make them well. Thanks for the patch!

PR #450, opened 2020-10-11
Increase timeout for two tests (fixes #449)

### LINKED ISSUE #449: Error in 'should work with different elm on PATH' if no other elm version is installed
When running the tests with 'npm link; npm run test-only', I get a failure for one of the --compiler tests in flags.js

> 
> --compiler
>       ✓ Should fail if the given compiler can't be executed (604ms)
>       1) Should work with different elm on PATH
>       ✓ Should work with local different elm (2104ms)
>     --watch
>       ✓ Should re-run tests if a test file is touched (3900ms)
> ...
>  1) flags
>       --compiler
>        Should work with different elm on PATH:
>     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it 
> resolves. (/home/frank/src/3rd_party/node-test-runner-upstream/tests/flags.js)

I assume this is because I have only one version of Elm installed; System info:

Elm version: 0.19.1
Ubuntu 20.04.1 LTS

Is this test supposed to fail if only one Elm version is installed? If yes, I'd suggest we display a helpful error message, something along the lines of "This is supposed to fail if you don't have additional Elm versions installed on your system"

frankschmitt:
Adds increased timeouts (5000 msec) to two tests that were failing on a semi-regular basis (at least for me)

### DIFF
--- a/tests/flags.js
+++ b/tests/flags.js
@@ -288,7 +288,7 @@ describe('flags', () => {
         'elm-test',
         '--compiler=different-elm',
         path.join('tests', 'Passing', 'One.elm'),
-      ]);
+      ]).timeout(5000);
 
       assert.equal(runResult.status, 0);
     });
--- a/tests/flags.js
+++ b/tests/flags.js
@@ -288,10 +288,10 @@ describe('flags', () => {
         'elm-test',
         '--compiler=different-elm',
         path.join('tests', 'Passing', 'One.elm'),
-      ]).timeout(5000);
+      ]);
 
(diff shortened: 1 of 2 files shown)


harrysarson:
CI failures look legitimate. I think yoh may have attached the timeout call to the wrong bit. Otherwise looks good!

frankschmitt:
What a stupid mistake. Sorry for that, fixed.

harrysarson:
Thanks!

Issue #441, opened 2020-09-24
Port CI tests to mocha


harrysarson:
We currently have some tests in https://github.com/rtfeldman/node-test-runner/blob/master/tests/ci.js that were originally a bash script and were ported to javascript some time ago. However, these tests do not use the mocha framework (unlike the other tests in <https://github.com/rtfeldman/node-test-runner/blob/master/tests>). 

If someone could take each test in the file (they are roughly split up by code like `
shell.echo('### Testing elm-test on example-application/');`) and put each test into the mocha `test("name", () => {})` format that would be great :) You may want to look at the tests in <https://github.com/rtfeldman/node-test-runner/blob/master/tests/flag.js> for inspiration.

There is some code in <https://github.com/rtfeldman/node-test-runner/blob/master/tests/ci.js> that uses `npm link` to make the global `elm-test` command run the code in your local repository. That can be left as is and mocha will run it before any tests. 

Drop a comment below if you are interested or have any questions!

frankschmitt:
I'd like to give this one a shot. Just to clarify: you want the existing ci.js replaced with the new version? And: how can I verify the npm link stuff still works?
(my Javascript / npm experience is quite limited, but I love Elm and I'd like to contribute)

harrysarson:
> you want the existing ci.js replaced with the new version? 

Exactly!

> how can I verify the npm link stuff still works?

The npm link stuff replaces the global `elm-test` exe with elm test from your local repository. (As if you published a new version of elm-test with your changes and then ran `npm install -g elm-test). The way to check it works is to modify the elm-test source code so that it crashes and make sure tests start failing!

harrysarson:
Removing help wanted to give @frankschmitt time to have a go!

lydell:
There are some nice improvements in ci.js here: https://github.com/rtfeldman/node-test-runner/pull/442/files#diff-669efeda79bb547b3537ab544fd28f75

What I did was to log more information when a test fails so it’s easier to understand what’s wrong.

It would be nice if the new mocha version can give helpful information on failures as well!

frankschmitt:
@lydell So would it make sense to implement the changes based on the version from your PR? Or rather on the vanilla version?

harrysarson:
[...]

PR #452, opened 2020-10-15
Port ci tests to mocha 441

frankschmitt:
First stab at #441. I tried to stick to the improved error messages from #442 - please let me know of any changes that might be required.


tests/ci.js
```diff
@@ -1,181 +1,204 @@
...
-  shell.exit(1);
-}
+describe('--help', () => {
+  it('Should exit indicating success', () => {
+    const runResult = execElmiToJSON(['--help']);
+    assertTestSuccess(runResult);
+  }).timeout(60000);
+
+  it('Should print the usage', () => {
+    const runResult = execElmiToJSON(['--help']);
```
harrysarson, line 72: also `execElmTest`!

tests/ci.js
```diff
@@ -1,181 +1,204 @@
...
+const fs = require('fs-extra');
+const os = require('os');
+const xml2js = require('xml2js');
+const readline = require('readline');
+const stripAnsi = require('strip-ansi');
+
+const { fixturesDir, spawnOpts, dummyBinPath } = require('./util');
+
+const elmiToJSONPath = require('elmi-to-json').paths['elmi-to-json'];
+const elmtestPath = path.join(__dirname, '..', 'bin', 'elm-test');
```
harrysarson, line 16: Nice! Is much better to be running these tests using local binary 👍

tests/ci.js
```diff
@@ -1,181 +1,204 @@
...
+  );
 }
 
-function assertTestFailure(testFile, clearCache) {
-  const result = run(testFile, clearCache);
-  if (result.code < 2) {
-    fail('Expected tests to fail with exit code 2 or higher.', result);
-  }
+function getDetailedMessage(message, runResult) {
+  return message + '\n\n' + 'STDOUT\n' + runResult.stdout + '\n\n' + 'STDERR\n' + runResult.stderr;
```
harrysarson, line 47: Nice 👍

harrysarson (commented):
Thanks @frankschmitt this looking great so far! I have just a couple of comments, be interested to hear what you think.

@lydell if you have the time, I would appreciate your eyes on this but no stress of course :)

harrysarson:
(also linting has failed, run `npm run format:write` to fix)

tests/ci.js
```diff
@@ -1,181 +1,204 @@
...
+describe('--version', () => {
+  it('Should exit indicating success', () => {
+    const runResult = execElmTest(['--version']);
+    assertTestSuccess(runResult);
+  }).timeout(60000);
+
+  it('Should print the usage', () => {
+    const runResult = execElmTest(['--version']);
+    assert.strictEqual(elmTestVersion, runResult.stdout.trim());
+  }).timeout(60000);
```
lydell, line 88: - Could these be merged into one? `Should print the version and exit with 0`.
- Does this really need a 1 minute timeout? If `elm-test --version` takes one minute we must have done something very wrong?

tests/ci.js
```diff
@@ -1,181 +1,204 @@
...
-shell.cd('../');
-
-shell.echo('\n### Testing elm-test on example-application-src/');
-
-shell.cd('example-application-src');
-
-assertTestSuccess('src');
+describe('Testing elm-test on an example application', () => {
+  before(() => {
+	  shell.pushd('example-application');
```
lydell, line 95: Is it possible to pass the `cwd` option to `spawn.sync` instead? Then we should be able to get rid of `shelljs` altogether?

tests/ci.js
```diff
@@ -1,181 +1,204 @@
...
-if (versionRun.code !== 0) {
-  shell.exec(
-    'echo Expected elm-test --version to exit with exit code 0, but it was ' +
-      versionRun.code
-  );
-  shell.exit(1);
-}
+describe('--help', () => {
+  it('Should exit indicating success', () => {
+    const runResult = execElmiToJSON(['--help']);
```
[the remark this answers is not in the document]
frankschmitt: To be completely honest, I wasn't 100% sure about this myself. Here's the snippet from the original:
```
shell.echo(filename + ': Verifying elmi-to-json is installed...');
var interfaceResult = spawn.sync(interfacePath, ['--help']);
var interfaceExitCode = interfaceResult.status;

if (interfaceExitCode !== 0) {
  shell.echo(
    filename +
      ': Failed because `elmi-to-json` is present, but `elmi-to-json --help` returned with exit code ' +
      interfaceExitCode
...
```
As far as I understand it, this simply verifies that elmi-to-json is present. Can we get rid of the elmi-to-json part altogether?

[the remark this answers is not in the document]
lydell: I think so, yes! Because:

- I don’t understand why it’s validated in the first place
- I’m removing elmi-to-json altogether in #442

lydell:
Don’t forget to run eslint as well! Here are my tips:

- Make sure ESLint and Prettier run in your editor. It’s super helpful! You get feedback on errors right in your editor and never have to format any code by hand.

- Run `npm test` locally sometimes. It’s faster than waiting for Travis.

tests/ci.js
frankschmitt: That's true - I took the timeout(60000); from flags.js, where it runs the tests for a whole package. I've set it to 5 seconds and merged both tests into one.

frankschmitt:
> Don’t forget to run eslint as well! Here are my tips:
> 
>     * Make sure ESLint and Prettier run in your editor. It’s super helpful! You get feedback on errors right in your editor and never have to format any code by hand.

Good suggestion. I just started using Visual Studio Code and still have a very basic configuration (normally, I prefer vim, but Javascript editing is super easy in VS Code).

> 
>     * Run `npm test` locally sometimes. It’s faster than waiting for Travis.

So far, I've always used  'npm run test-only', but 'npm test' includes a couple of additional useful stuff. Nice!

frankschmitt:
> Don’t forget to run eslint as well! Here are my tips:

This raises a small problem. eslint is complaining about this line:

const { fixturesDir, spawnOpts, dummyBinPath } = require('./util');

because dummyBinPath is never used; the error message is:

  9:33  error  'dummyBinPath' is assigned a value but never used  no-unused-vars

I tried to circumvent this by prefixing the variable name with '_', but this doesn't work. I guess the most sensible approach is to set the varsIgnorePattern as described in the [docs](https://eslint.org/docs/rules/no-unused-vars#varsignorepattern). I've got two questions regarding this:
- where do I need to set this?
- what would be a sensible pattern? "ignored", everything starting with _ or ... ?

lydell:
If `dummeBinPath` is unused, can’t you change the code to:

```
const { fixturesDir, spawnOpts } = require('./util');
```

?

tests/ci.js
harrysarson: This would be really cool. I love to reduce the number of deps we take. Feel free to leave this for another pull request though if you want.

tests/ci.js
frankschmitt: > If `dummyBinPath` is unused, can’t you change the code to:
> 
> ```
> const { fixturesDir, spawnOpts } = require('./util');
> ```
> 
> ?

I had assumed that the number of assignment targets has to match the number of exports from util.js. Fixed, just another indicator of my very limited Javascript knowledge :-)

tests/ci.js
```diff
@@ -1,181 +1,204 @@
...
+    shell.popd();
+  });
+  
+  it('Should succeed for the passing tests', () => {
+    shell.ls('tests/Passing/').forEach(function (testToRun) {
+      const itsPath = path.join('tests', 'Passing', testToRun);
+      const runResult = execElmTest([itsPath]);
+      assertTestSuccess(runResult);
+    });
+  }).timeout(60000);
```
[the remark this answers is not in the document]
frankschmitt: Sounds good, I'll make the necessary changes to the code for example-application, example-package and the tests for single Elm files.

tests/ci.js
```diff
@@ -1,181 +1,204 @@
...
-  let cmd = [elmTest, ...(testFile ? [testFile] : []), '--color'].join(' ');
 
-  shell.echo('Running: ' + cmd);
-  return shell.exec(cmd, spawnOpts);
+function execElmiToJSON(args) {
+  return spawn.sync(
+    elmiToJSONPath,
+    args,
+    Object.assign({ encoding: 'utf-8' }, spawnOpts)
+  );
```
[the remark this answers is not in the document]
frankschmitt: Done.

tests/ci.js
frankschmitt: It's changed.

tests/ci.js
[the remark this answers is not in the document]
frankschmitt: I've removed everything related to elmi-to-json.

tests/ci.js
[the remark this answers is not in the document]
frankschmitt: On second thought, I think I'll leave example-application and example-packages as-is - they only have a single file for failing / passing tests each, so it doesn't really make sense to switch (unless we want to split the TestsFailing.elm / TestsPassing.elm files, but this seems overkill to me.

frankschmitt:
> If `dummyBinPath` is unused, can’t you change the code to:
> 
> ```
> const { fixturesDir, spawnOpts } = require('./util');
> ```
> 
> ?

I had assumed that the number of assignment targets has to match the number of exports from util.js. Fixed, just another indicator of my very limited Javascript knowledge :-)

tests/ci.js
frankschmitt: Currently, we're also using shelljs to enumerate test files (using shell.ls). What would be a good alternative to that (without introducing another dependency) ?

lydell: Something like this should work:

```js
    const dir = path.join(fixturesDir, 'tests', 'Passing');
    fs.readdirSync(dir).forEach(function (testToRun) {
      const itsPath = path.join(dir, testToRun);
      const runResult = execElmTest([itsPath]);
      assertTestSuccess(runResult);
    });
```

Where `fs` comes from `const fs = require('fs')` which is the [native Node.js fs module](https://nodejs.org/api/fs.html).

tests/ci.js
```diff
@@ -1,181 +1,243 @@
...
+      const itsPath = path.join('tests', 'Failing', testToRun);
+      const runResult = execElmTest([itsPath]);
+      assertTestFailure(runResult);
+    }).timeout(5000);
+  }
 
-/* ci tests on single elm files */
+  it(`Should run every file in tests/Failing`, () => {
+    const filesFound = fs.readdirSync('tests/Failing/');
+    assert.deepStrictEqual(filesFound, failingTestFiles);
```
harrysarson, line 225:
```suggestion
    filesFound.sort();
    assert.deepStrictEqual(filesFound, failingTestFiles);
```

I do not think `readDirSync` guarantees the order if the values it returns.

tests/ci.js
```diff
@@ -1,181 +1,243 @@
...
-shell.ls('tests/RuntimeException').forEach(function (testToRun) {
-  shell.echo(
-    '\n### Testing ' +
-      testToRun +
-      ' (expecting it to error with a runtime exception)'
-  );
-  assertTestErrored(path.join('tests', 'RuntimeException', testToRun));
+  it(`Should run every file in tests/RuntimeException`, () => {
+    const filesFound = fs.readdirSync('tests/RuntimeException/');
+    assert.deepStrictEqual(filesFound, erroredTestFiles);
```
harrysarson, line 241: Ditto

tests/ci.js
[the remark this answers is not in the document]
harrysarson: Modulo by sorting comment, it looks perfect now. Nice job 👍

harrysarson:
Looking really good!

tests/ci.js
lydell: Why are we asserting what files are in the fixtures folders?

frankschmitt: It was a suggestion by @harrysarson to avoid accidentally not running tests : 

> and if somehow we mess around with the directory structure so that tests are not found by shells.ls(...) then we get an error rather than silently not running the tests.

This sounds sensible to me, but I don’t have strong opinions either way.

harrysarson: Yeah, I am thinking it would be very easy to add a file to the test fixtures directory and just assume it is tested. If this test fails we get an error message which should act as a fairly good pointer to the array you have to add your test case name to.

tests/ci.js
frankschmitt: I've removed shelljs (cwd is now provided to spawn.sync, and we use fs.readdirSync() instead of shell.ls).

tests/ci.js
```diff
@@ -18,11 +17,13 @@ const resultSuccess = 0;
 const resultErrored = 1;
 const resultFailureThreshold = 2;
 
-function execElmTest(args) {
+function execElmTest(args, cwd) {
+  // default: current directory
+  cwd = typeof cwd !== 'undefined' ? cwd : '.';
```
lydell, line 22: I think this should work:

```suggestion
function execElmTest(args, cwd = '.') {
```

harrysarson: Will this work in node 8?

frankschmitt: In fact, I first tried it with the default argument syntax (cwd = '.'), but I got test failures - apparently, this syntax was introduced in ES 2015 (tbh, I don't know which version node 8 is based on). So I guess we're stuck with the more cumbersome approach (cwd = typeof ...) for the time being.

harrysarson:
@frankschmitt is this ready to merge from your point of view?

tests/ci.js
harrysarson: Dropping node 8 support is on our short term road map. We can fix this once we have done that.

frankschmitt:
Yes, from my point of view it‘s ready to merge.

harrysarson:
Thank you, this is a really nice improvement to the test suite.

Issue #399, opened 2019-11-19
elm-test --watch crashes if Elm code doesn't compile


MartinSStewart:
Steps to reproduce:
1. Download [sscce.zip](https://github.com/rtfeldman/node-test-runner/files/3863113/sscce.zip) and install NPM packages
2. Run `yarn test` in terminal
3. Notice that elm-test will start and then immediately crash with the following exception
```
(node:3644) UnhandledPromiseRejectionWarning: Compilation failed while attempting to run `elm make` on /Users/martinstewart/Desktop/sscce
(node:3644) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:3644) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

Node version: v10.16.0
OS: MacOS
Elm version: 0.19.1

lydell:
Update! It used to crash like this:

```
-- CYCLIC DEFINITION ------------------------------------------ src/MyModule.elm

The `a` value is defined directly in terms of itself, causing an infinite loop.

4| a = a
   ^
Are you are trying to mutate a variable? Elm does not have mutation, so when I
see a defined in terms of a, I treat it as a recursive definition. Try giving
the new value a new name!

Maybe you DO want a recursive value? To define a we need to know what a is, so
let’s expand it. Wait, but now we need to know what a is, so let’s expand it...
This will keep going infinitely!

Hint: The root problem is often a typo in some variable name, but I recommend
reading <https://elm-lang.org/0.19.1/bad-recursion> for more detailed advice,
especially if you actually do need a recursive value.

(node:93627) UnhandledPromiseRejectionWarning: Compilation failed while attempting to run `elm make` on /Users/lydell/Downloads/sscce
(Use `node --trace-warnings ...` to show where the warning was created)
(node:93627) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
[...]

PR #451, opened 2020-10-13
Leverage `elm-json solve` (updated)

lydell:
Fixes #328, fixes #277, fixes #273, fixes #323, fixes #374

This is #356 but merged with `master`.

I also noticed that the return value of `Runner.getIndirectDeps` was ultimately unused, so I removed that as well. That means we no longer use `elmi-to-json` to translate package dependency ranges to concrete versions! `elm-json` takes care of it all (I didn’t even need to add any code for it, @zwilias’ code already took care of it).

Paired with #442 we can get rid of `elmi-to-json` altogether.

PR #453, opened 2020-10-15
Fix watch for packages when there are compilation errors in src/

lydell:
Fixes https://github.com/rtfeldman/node-test-runner/issues/399.

Previously, if there was an error in `src/` for a _package_ project, `elm-test --watch` would immediately exit. This was because we compile `src/` first to get .elmi files for the files in there so that `elmi-to-json` can find the exact dependencies there. But since #451 we don’t use `elmi-to-json` to find exact dependencies anymore – we use `elm-json` instead – so there’s no need for that compilation step anymore. I removed that, which fixes the issue and also improves test run time since the compiler is invoked one time less.

As for the `undefined` log issue, that was because of promises being rejected with strings while the `.catch` code expected `Error` objects. `"some string".message` is undefined, so that’s where it came from. I updated to reject with `Error` objects instead.

This allowed for some pretty nice cleanups.


lib/Compile.js
```diff
@@ -94,13 +94,7 @@ function compileSources(
...
         resolve();
       } else {
-        const msg =
-          'Compilation failed while attempting to ' +
-          (testFilePaths.length > 0
-            ? 'build ' + testFilePaths.join(' ')
-            : 'run `elm make` on ' + projectRootDir);
-
-        reject(msg);
+        reject(new Error(`\`elm make\` failed with exit code ${exitCode}.`));
```
lydell, line 97: The old error was super verbose and not useful when the nice error from `elm make` is printed just above anyway.

lib/elm-test.js
```diff
@@ -305,77 +304,53 @@ function infoLog(msg) {
...
-  // Ergo, globify all the arguments we receive.
-  const fileGlobs = files.length > 0 ? files : [];
-  const testFilePaths = resolveGlobs(files);
-  const hasBeenGivenCustomGlobs = fileGlobs.length > 0;
-  const elmJsonPath = path.resolve(path.join(projectRootDir, 'elm.json'));
+try {
+  projectElmJson = fs.readJsonSync(elmJsonPath);
+} catch (err) {
+  console.error('Error reading elm.json: ' + err.message);
+  throw process.exit(1);
```
harrysarson, line 325: Do we have to crash if we are in watch mode and the project elm json is invalid?

harrysarson: And what happens if someone edits their elm.json while we are watching tests>

harrysarson (approved):
Looking good. I always love to see patches with more lines removed than added.

lib/elm-test.js
lydell: I didn’t think about that case! This could for sure be a future enhancement.

Currently this is done outside of the watch loop so it would require more changes than I’m comfortable with at the moment. This _looks_ like added code but it’s in fact just moved so were keeping the same behavior as before regarding elm.json at least.

harrysarson: Before this PR `generateElmJson` read the elm.json file from disk everything. Would that mean that it would detect changes in your elm.json during watch mode?

lydell: Ah, good catch, didn’t think about that change. I’ll have to investigate.

lydell: Turns out `Generate.generateElmJson` is outside the watch loop too (and was before this PR as well). And moving it into the loop brings more issues:

- We don’t watch `elm.json` so you need to edit some .elm file to trigger.
- If you edit `"source-directories"` we need to recalculate which globs the watcher should look at, which is also outside the loop.
- Reading elm.json and running `elm-json` currently does `process.exit(1)` on failure which would need to be refactored.
- We basically need to restart the whole thing if elm.json changes.

My conclusion is:

- This PR has not changed behavior.
- Adding support for elm.json changes could be a nice future improvement.

harrysarson: > My conclusion is:
> 
> * This PR has not changed behavior.

Good enough for me!

PR #455, opened 2020-10-17
Remove unnecessary fixtures

lydell:
Let’s see what Travis says about this … might need to cache `~/.elm`

… ok build succeeded in about the same time as usual, let’s try caching and see what happens

… first run with cache took a couple of minutes more (lots of cache uploads, I guess), let’s see how a run with a cache hit (hopefully!) goes

… Travis indicates cache hits, but the runs take the same time though

Maybe downloading Elm packages is super fast, so the caching is more of a reliability thing thing if package.elm-lang.org is down?

---

Ok, so in the end this PR became three things:

- Remove unnecessary fixtures
- Run Flow and ESLint in Travis
- Cache ~/.elm in Travis

Those aren’t really related. Let me know if you want me to split this PR up!

Also, I’m not sure about the caching if we want it or not. (Btw, the AppVeyor docs on caching are a bit scary – is it possible to run out of cache space? https://www.appveyor.com/docs/build-cache/#cache-size-beta)


harrysarson:
To clarify this PR changes `.travis.yml`, `.package.json` and deletes all generated fixture files?

lydell:
Yes, more or less!

```
❯ git diff master --stat=9999999 | grep -v tests/fixtures/elm-home
 .travis.yml                                                                                                          |    9 +++++++--
 package.json                                                                                                         |    3 ++-
 tests/fixtures/empty                                                                                                 |    0
 tests/util.js                                                                                                        |    3 +--
 952 files changed, 10 insertions(+), 169660 deletions(-)
```

And here’s the diff for those files (`tests/fixtures/empty` was an empty file that is now removed):

```
❯ git --no-pager diff master .travis.yml package.json tests/util.js
diff --git a/.travis.yml b/.travis.yml
index 5b4c555..8a9e6f8 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,11 @@ script:
 jobs:
   include:
     - stage: other
-      script: npm run format:check
+      script: npm run check
       node_js: lts/*
-      name: Lint
+      name: Check
+
+cache:
+  npm: true
+  directories:
+    - $HOME/.elm
diff --git a/package.json b/package.json
index 1c48478..f48564b 100644
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
     "flow": "flow",
     "lint": "eslint .",
     "test-only": "mocha tests && cd elm && node ../bin/elm-test",
-    "test": "flow check && npm run lint && npm run format:check && npm run test-only",
+    "test": "npm run check && npm run test-only",
+    "check": "flow check && npm run lint && npm run format:check",
     "format:check": "prettier --check . && elm-format elm --validate",
     "format:write": "prettier --write . && elm-format elm --yes"
   },
diff --git a/tests/util.js b/tests/util.js
index 0d5f505..cfead24 100644
--- a/tests/util.js
+++ b/tests/util.js
@@ -2,12 +2,11 @@ const path = require('path');

 const fixturesDir = path.join(__dirname, 'fixtures');

-const elmHome = path.join(fixturesDir, 'elm-home');
 const dummyBinPath = path.join(fixturesDir, 'dummy-bin');
 const newPath = process.env.PATH + path.delimiter + dummyBinPath;
 const spawnOpts = {
   silent: true,
-  env: Object.assign({ ELM_HOME: elmHome }, process.env, {
+  env: Object.assign({}, process.env, {
     PATH: newPath,
     Path: newPath,
   }),
```

PR #456, opened 2020-10-17
Avoid recalculating dependencies if not needed

lydell:
Extracted from #442.

Running `elm-json solve` takes a second or so. But projects’ elm.json rarely change so we shouldn’t need to calculate dependencies every single time we run elm-test.

This PR creates an elm.template.json next to our generated elm.json which we can compare between runs to see if we need to solve dependencies again or not. See the code comments for more details.

I pretty consistently get 1 second faster test runs in various projects I try PR on.

### DIFF
--- a/lib/Generate.js
+++ b/lib/Generate.js
@@ -174,7 +174,7 @@ function generateElmJson(
   }
 
   let testElmJson = Object.assign({}, testElmJsonTemplate, {
-    dependencies: Solve.get_dependencies(elmJsonPath, projectElmJson),
+    dependencies: Solve.get_dependencies(elmJsonPath),
     'test-dependencies': {
       direct: {},
       indirect: {},
--- a/lib/Generate.js
+++ b/lib/Generate.js
@@ -60,7 +60,6 @@ function generateElmJson(
   const generatedSrc = path.join(generatedCodeDir, 'src');
 
   var isPackageProject = projectElmJson.type === 'package';
-  var dependencies = Solve.get_dependencies(elmJsonPath);
 
   // if we were given file globs, we don't need to check the tests/ directory exists
   // this is only for elm applications, which people may need to introduce slowly into their apps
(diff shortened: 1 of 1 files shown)


lib/Generate.js
```diff
@@ -152,9 +138,51 @@ function generateElmJson(
...
+      // …then just return early and go with the generated elm.json we already have.
+      // This avoids calculating exact dependencies again and saves a second or so.
+      return [generatedCodeDir, generatedSrc, sourceDirs];
+    }
+  } catch (_error) {
+    // No template file since before, moving on.
+  }
+
+  let testElmJson = Object.assign({}, testElmJsonTemplate, {
+    dependencies: Solve.get_dependencies(elmJsonPath, projectElmJson),
```
harrysarson, line 177: Does `Solve.get_dependencies` not only take a single argument? https://github.com/rtfeldman/node-test-runner/blob/master/lib/solve.js#L3

harrysarson (commented):
I do like this. I am a bit cautious because clever caching methods can easily lead to really trick bugs but if we are careful I can see how this would be a really nice perf boost.

Could we do mtime based caching, the way that `make` caches things. I am thinking something like:

```
if (stat -c %Y {generatedCoreDir}/elm.json) > (stat -c %Y {elmJsonPath}) then
  Solve to get new test elm.json
else
  Use existing elm.json
```

In my mind this would be slightly more error resistant than generating a new different file for cache invalidation.

lib/Generate.js
lydell: Correct, bad extraction from #442 (it had two arguments in an experiment over there)

lydell:
Caching things sure is scary! Super easy to get it wrong.

First, just to check if we’re on the same page here. The condition is flipped in your pseudo-code, isn’t it? Here’s how I envision the mtime check:

```
if project elm.json is newer than generated elm.json then
	solve to get new generated elm.json for sure
else
	we actually don’t know if we need to update generated elm.json or not
```

Why don’t we know if we need to update? Because the generated elm.json does not _only_ depend on the project elm.json as input. It also depends on whether the `tests/` directory exists and if you have passed globs to `elm-test` – this affects whether we should include `tests/` as a source-directory or not.

I suppose we could do this:

```
if project elm.json is newer than generated elm.json then
	solve to get new generated elm.json
else
	calculate source dirs
	read generated elm.json
	if calculated source dirs are the same as in generated elm.json then
		use existing generated elm.json
	else
		solve to get new generated elm.json
```

I’ve always heard mtime checks have their own problems and that you should avoid them if possible. Here’s one blog post about it: https://apenwarr.ca/log/20181113

For example, it’s possible that your project elm.json ends up with changes _and_ an old mtime. Then elm-test will do the wrong thing. Whereas my content based approach wouldn’t. And even the mtime version requires a content based check for source-directories so why not go all the way?

Here’s my take:

- The content based approach feels more robust and easier to implement in this case!
- The speedup from this PR is noticeable and really nice so I think it’s worth the risk caching always brings!

harrysarson:
> I’ve always heard mtime checks have their own problems and that you should avoid them if possible. Here’s one blog post about it: https://apenwarr.ca/log/20181113

Thanks for the link! Let's drop the mtime idea then!

harrysarson:
> Why don’t we know if we need to update? Because the generated elm.json does not only depend on the project elm.json as input. It also depends on whether the tests/ directory exists and if you have passed globs to elm-test – this affects whether we should include tests/ as a source-directory or not.

Is this true:

```
generatedElmJson.dependancies  --- depends on --- sourceJson.dependancies
generatedElmJson               --- depends on --- sourceJson.dependancies --- and --- whether tests/ exists and globs
```

If so could our "cache invalidation file" contain both `sourceJson.dependancies` and `generatedElmJson.dependancies`? Then we *always* regenerate `generatedElmJson` but use the cached dependancies when valid. Using the template as the "cache invalidation file" feels slightly not right to me although I am not doing a great job of articulating why,

harrysarson:
This work feels like a prerequisite for properly handing changes to `elm.json` files in watch mode. Therefore, it is very exciting and useful.

lydell:
The data depends like so:

```
generatedElmJson.dependencies --- depends on sourceJson.dependencies
generatedElmJson.source-directories --- depends on sourceJson.source-directories + whether tests/ exists + whether globs given
```

So in theory would even could do partial updates of the generated elm.json (haven’t even thought about that before).

I think your idea can work – I’ll try and implement it as a separate PR so we can compare.

lydell:
Closing in favor of #457

PR #457, opened 2020-10-18
Avoid recalculating dependencies if not needed – cache elm-json outputs

lydell:
This does the same as #456 but using another approach. We _always_ generate a new elm.json, but using cached results of `elm-json` if possible. This saves a second or so.

Advantages over #456:

- Much easier to understand.
- If the user edits their source-directories (but not their dependencies) that no longer invalidates the cache.


lib/Solve.js
```diff
@@ -0,0 +1,67 @@
...
+
+  try {
+    return fs.readJsonSync(cacheFile);
+  } catch (_error) {
+    // Cache file does not exist or is malformed. Move on.
+  }
+
+  const dependencies = getDependencies(elmJsonPath);
+
+  fs.writeFileSync(cacheFile, JSON.stringify(dependencies));
```
harrysarson, line 34: We `JSON.parse` and then immediately `JSON.stringify`. We can save ourselves some work here I think.

lib/Solve.js
```diff
@@ -0,0 +1,67 @@
...
+  projectElmJson /*: any */
+) /*: { direct: { [string]: string }, indirect: { [string]: string } } */ {
+  const hash = sha1(
+    JSON.stringify({
+      dependencies: projectElmJson.dependencies,
+      'test-dependencies': projectElmJson['test-dependencies'],
+    })
+  );
+
+  const cacheFile = path.join(generatedCodeDir, `dependencies.${hash}.json`);
```
harrysarson, line 24: I really like this! 

I only wonder if there is a way of cleaning up old cache files so that we don't use up too much disk space.

harrysarson (commented):
I like this method a lot!

harrysarson:
CI failures are `solve.js` vs `Solve.js` problems I think. MacOs' case insensitive filesystem is causing errors here.

lib/Solve.js
lydell: Do we need to worry about that?

- The files are very small (< 4 kB).
- elm.json changes very seldomly so there won’t be many versions.
- We don’t clean up anything else: I currently have 63 MainXXXXXXX.elm generated by elm-test, and also old generated code for previous versions of elm-test. This takes ~2 MB of space (not counting the nested elm-stuff folders).

harrysarson: Wow, okay. That is a lot of MainXxxxxx.elm! I think we are probably okay then.

lib/Solve.js
```diff
@@ -0,0 +1,66 @@
...
+
+  try {
+    return fs.readJsonSync(cacheFile);
+  } catch (_error) {
+    // Cache file does not exist or is malformed. Move on.
```
harrysarson, line 29: One last question. What are you thoughts on checking whether `error.code === 'ENOENT'` and otherwise rethrowing the error? I guess it boils down to: is it better to sometimes crash when we could actually recover or to (sometime in the hypothetical future) start failing to cache elm.json dependencies and not notice?

lydell: Very good question.

- Current approach: elm-test “self-heals” from a bad file, but on the other hand we might miss legit errors that we would have wanted to know about and code around.

- Your approach: elm-test fails hard on a bad file. Hopefully the user reports it to use. User needs to remove the file (or the whole `elm-stuff`) to get going again. We could mention all of this in the error message, though. Now that I think of it – that sounds like the best solution.

harrysarson: Or we print a warning to stderr if there is a non `error.code === 'ENOENT'` exception but do not crash.

lydell: I’m printing a warning now. Sample:

```
Ignoring bad dependencies cache file:

/Users/lydell/forks/node-test-runner/example-package/elm-stuff/generated-code/elm-community/elm-test/0.19.1-revision4/dependencies.d48585830c2c9ad9efb7f8be7173b55e9509bafd.json: Unexpected end of JSON input

Please report this issue: https://github.com/rtfeldman/node-test-runner/issues/new

elm-test 0.19.1-revision4
-------------------------

Running 16 tests. To reproduce these results, run: elm-test --fuzz 100 --seed 227195913342066

↓ TestsFailing
↓ basic expectations
…
```

harrysarson (approved):
:heart:

Issue #446, opened 2020-10-03
Add elm review for our elm code


harrysarson:
I cannot write instructions for this as I have never used elm-review before! First step is to workout the instructions and define our goals here.

It would be great to ensure the highest quality of our elm code.

jfmengels:
I recommend starting with enabling the rules from [`elm-review-unused`](https://package.elm-lang.org/packages/jfmengels/elm-review-unused/latest).

You can start with
- Adding the `elm-review` CLI to the probject: `npm install -D elm-review`
- Adding a configuration. A simple way to do this is by running `elm-review init --template jfmengels/elm-review-unused/example` in the `elm` root of the project.
- Running it in the test suite by running: It's just `elm-review` (I don't know where you prefer doing this)

Then you'll have to fix issues. `elm-review --fix` or `elm-review --fix-all` can help you out. `--fix-all` can be very slow if there are are a lot of issues, so it might not be the best start.

If there are too many issues, the PR diff will be very big, so it might be a pain for @harrysarson to review. I would recommend to take the configuration and comment out every rule except `NoUnused.Variables` and make the PR.

It DOES look like there are not too many of these issues, so running `--fix-all` is just fine. That said, a lot of the errors are about unused functions in `Console.Text`, maybe you want to ignore some rules for some rules.

Know that you can ignore errors using [ignoreErrorsForFiles](https://package.elm-lang.org/packages/jfmengels/elm-review/latest/Review-Rule#ignoreErrorsForFiles) and  [ignoreErrorsForDirectories](https://package.elm-lang.org/packages/jfmengels/elm-review/latest/Review-Rule#ignoreErrorsForDirectories).

Once that is set up, you can take a look at adding new rules. I recommend searching through Elm packages by searching for `elm-review` and finding packages (or just looking at the ones I made). Hopefully with Hacktoberfest we'll be able to add a lot of documentation-related rules.

I hope this helps!

frankschmitt:
I'd be willing to give this one a go (I've never used elm-review, but this seems like the perfect opportunity to get acquainted with it).

frankschmitt:
@harrysarson I'd propose a staged approach for this (essentially what @jfmengels suggested, a little bit more formalized)
1. Establish elm-review support
   * provide basic configuration for all Elm subdirectories (using only the NoUnused.Variables rule)
2. Automate & integrate into CI
   * figure out how to integrate elm-review into
      ** npm test
      ** Travis
      ** Appveyor
[...]

PR #458, opened 2020-10-18
Establish elm review 446

frankschmitt:
First try at #446. I'm not 100% happy with the config for the subdirs - e.g. example-package-no-tests has just a single source file, and I had to exclude this file from the single rule I've enabled so far, so elm-review for this is essentially a no-op. 
Feedback / suggestions welcome :-)


harrysarson:
Really nice thank you! I think we probably do not want to elm-review on our examples (we should keep them as simple as possible to emphasise the testing bits) which would also sidestep the subdir issue.

frankschmitt:
> I think we probably do not want to elm-review on our examples (we should keep them as simple as possible to emphasise the testing bits) which would also sidestep the subdir issue.

Sounds good. I've removed the elm-review stuff for the examples.

package.json
```diff
@@ -58,6 +59,7 @@
   "devDependencies": {
     "elm": "0.19.1-3",
     "elm-format": "0.8.3",
+    "elm-review": "^2.3.3",
```
harrysarson, line 62: Can we pin this dep:

```suggestion
    "elm-review": "2.3.3",
```

harrysarson (approved):

PR #461, opened 2020-10-24
Drop support for Node.js 8

lydell:
This drops support for Node.js 8, bumping the minimum supported version to v10.13.0 (released 2018-10-30). I chose that particular version rather than v10.0.0 because:

- That’s the oldest version of the 10.x branch marked as LTS (see https://nodejs.org/en/download/releases/).
- This allows us to use `fs.mkdirSync(dir, { recursive: true })` (recursive was added in Node.js v10.12.0 – see https://nodejs.org/api/fs.html#fs_fs_mkdirsync_path_options). This way Node.js has all we need to get rid of the `fs-extra` dependency.

I also added Node.js 15 to CI.

This is required by #442, which needs to use the “unicode property escapes” feature of regex which is supported only in Node.js 10+ (https://github.com/rtfeldman/node-test-runner/pull/442#discussion_r495579452).

The idea is that we should be able to release this as version 0.19.1-revision5. It’s technically a breaking change, but Node.js 8 has been end of life since 2019-12-31 and our versioning scheme doesn’t really have space for marking this as a breaking change. People who absolutely can’t upgrade Node.js can stay on an older version of node-test-runner until they’ve figured out an upgrade path.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2216,9 +2216,9 @@
       "integrity": "sha1-Ywn04OX6kT7BxpMHrjZLSzd8nns="
     },
     "picomatch": {
-      "version": "2.2.1",
-      "resolved": "http://verdaccio.internal.utterberry.com/picomatch/-/picomatch-2.2.1.tgz",
-      "integrity": "sha512-ISBaA8xQNmwELC7eOjqFKMESB2VIqt4PPDD0nsS95b/9dZXvVKOlz9keMSnoGGKcOHXfTvDD6WMaRoSc9UuhRA=="
+      "version": "2.2.2",
+      "resolved": "https://registry.npmjs.org/picomatch/-/picomatch-2.2.2.tgz",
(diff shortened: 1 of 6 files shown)


package-lock.json
```diff
@@ -2216,9 +2216,9 @@
       "integrity": "sha1-Ywn04OX6kT7BxpMHrjZLSzd8nns="
     },
     "picomatch": {
-      "version": "2.2.1",
-      "resolved": "http://verdaccio.internal.utterberry.com/picomatch/-/picomatch-2.2.1.tgz",
```
lydell, line 2220: verdaccio.internal.utterberry.com? 😕

harrysarson: Yeah sorry, I once did something on my work pc and our internal package registry snuck into the package lock.

harrysarson:
Thanks!

package-lock.json
lydell: Ah, that explains it! I was worried this was something malicious or something

PR #462, opened 2020-10-24
Remove shelljs

lydell:
We don’t need shelljs in our tests – fs/fs-extra already has all we need, and we can use the `cwd` option when spawning processes.

(I intend to get rid of fs-extra throughout the entire code base in a future PR.)


harrysarson:
> (I intend to get rid of fs-extra throughout the entire code base in a future PR.)

fs-extra does something clever when we run out of file handles (?) We would want to find out what this is and whether we need it for the test runner before removing the dep. Otherwise, happy to see it go.

harrysarson:
Thanks!

PR #466, opened 2020-10-31
Add more review rules 446

frankschmitt:
Enables the remaining Unused review rules as discussed in #446. Things to note:
- a couple of files have been excluded because they're reported, but are essential for the executable to work properly (false positives, see discussion for #446)
- some minor changes to the copied external modules were necessary to make elm-review happy; if we decide to re-import them in the future, we'll need to re-apply these changes. This is not the optimal solution; we should probably either try to get the upstream modules to also use elm-review and clean up unused imports etc. or exclude them from elm-review.

Suggestions / feedback welcome :-)

### DIFF
--- a/elm/review/src/ReviewConfig.elm
+++ b/elm/review/src/ReviewConfig.elm
@@ -24,13 +24,20 @@ import Review.Rule exposing (Rule)
 
 config : List Rule
 config =
-    [ {- NoUnused.CustomTypeConstructors.rule []
-              , NoUnused.Exports.rule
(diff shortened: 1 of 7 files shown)


harrysarson:
Thank you!

> - some minor changes to the copied external modules were necessary to make elm-review happy; if we decide to re-import them in the future, we'll need to re-apply these changes. This is not the optimal solution; we should probably either try to get the upstream modules to also use elm-review and clean up unused imports etc. or exclude them from elm-review.

Let's worry about this when it is time to update these modules

PR #469, opened 2020-11-02
Remove unnecessary fixtures folder

lydell:


harrysarson:
I thought these were gone in #455 ?

lydell:
Turns out there was another one. #455 removed `tests/fixtures/`, this one `fixtures/` at the root. I believe all of them used to be in the root `fixtures/` but were moved to `tests/fixtures/` at some point, but some got left behind or accidentally committed later.

harrysarson:
Thanks!

PR #465, opened 2020-10-30
Improve elm-test.js

lydell:
Fixes #211 
Addresses https://github.com/rtfeldman/node-test-runner/pull/442#discussion_r495579938

This PR makes a bunch of improvements to elm-test.js:

- Properly validate CLI arguments:

  - Error on unknown flags.
  - Validate flag values early (seed and fuzz must be numbers, report must be console/json/junit).
  - Default flag values early, instead of in code generation or even in Elm.

- elm-test.js is now 100% definitions, except the last line which is `main();` to kick things off. Previously, it was difficult to understand in what order things executed. There was some top-level code, a bunch of function definitions, some more top-level code, more functions, and so on. Now, you can follow the call stack from `main()`. Also, this means that there are way fewer global variables – functions now get passed what they depend on.

- Use async/await where it helps (in preparation for merge conflicts with #464 – it should be possible to just pick this PR’s code for every conflict in elm-test.js).

- Improved help message. Also, add some aliases, such as `-h`, for showing help. I find it super annoying when CLI tools make it difficult to find the help – which is the key to all other commands and flags.

- Various cleanups.

Help before:

```
Usage: elm-test init # Create example tests

Usage: elm-test install PACKAGE # Like `elm install PACKAGE`, except it installs to "test-dependencies" in your elm.json

Usage: elm-test TESTFILES # Run TESTFILES, for example tests/**/*.elm

Usage: elm-test [--compiler /path/to/compiler] # Run tests

Usage: elm-test [--seed integer] # Run with initial fuzzer seed

Usage: elm-test [--fuzz integer] # Run with each fuzz test performing this many iterations

Usage: elm-test [--report json, junit, or console (default)] # Print results to stdout in given format

Usage: elm-test [--version] # Print version string and exit

Usage: elm-test [--watch] # Run tests on file changes
```

Intermediate help (outdated):

```
elm-test init
    Create example tests

elm-test
    Run tests in the tests/ folder

elm-test TESTFILES
    Run TESTFILES, for example "src/**/*Tests.elm"

elm-test install PACKAGE
    Like `elm install PACKAGE`, except it installs to
    "test-dependencies" in your elm.json

Options:
    --compiler /path/to/compiler
        Use a custom path to an Elm executable. Default: elm
    --seed INT
        Run with a previous fuzzer seed. Default: A random seed
    --fuzz INT
        Run with each fuzz test performing this many iterations. Default: 100
    --report json
    --report junit
    --report console
        Print results to stdout in the given format. Default: console
[...]

PR #470, opened 2020-11-15
Fix broken install command

lydell:
Regression from #465.

Before:

```
❯ ../bin/elm-test install truqu/elm-md5
I am having trouble with this argument:

    t

It is supposed to be a <package> value, like one of these:

    elm/html
    elm/http
    elm/svg
    elm/time
```

After:

```
❯ ../bin/elm-test install truqu/elm-md5
Here is my plan:

  Add:
    truqu/elm-md5            1.1.0
    zwilias/elm-utf-tools    2.0.1

Would you like me to update your elm.json accordingly? [Y/n]:
Success!
```

This happened because I thought the first argument of the `.action` callback is always an array. Wrong! The signature depends on the `.command` syntax provided:

```js
program
  .command('name <one> <two> [three] [rest...]')
  .action((one: string, two: string, three: string | undefined, rest: Array<string>, cmd: Command) => {
    // Do stuff
  });
```

For the install command we have:

```js
program
  .command('install <package>')
  .action((packageName: string, cmd: Command) => {})
```

Before we tried to destructure the first argument (thinking it was an array), so we only got the first letter of the package name.

Unfortunately our test suite doesn’t catch this.


harrysarson:
Nice catch! Could you add a regression test?

lydell:
Do you have any ideas on how to write one?

lydell:
I figured it out.

harrysarson:
It looks like travis has pulled macos support. I guess we cannot complain too much because we were getting it for free.

@Lydell could you remove mac from travis just so we can get a green tick. Longer time we probably want to migrate to github actions.

Cc: @rtfeldman 
Refs: https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

PR #472, opened 2020-11-15
Simplify test variant definition regex

lydell:
This regex is more specific than it needs to be, as mentioned here: https://github.com/mpizenberg/elm-test-rs/pull/15/files#r515490848

This PR simplifies it, making the regex somewhat easier to understand and makes us rely on ever so slightly fewer implementation details.


lib/Generate.js
```diff
@@ -44,7 +44,7 @@ function prepareCompiledJsFile(
 // variant of the `Test` type. To avoid having to update this regex if a new
 // variant is added, newer versions of elm-explorations/test have prefixed all
 // variants with `ElmTestVariant__` so we can match just on that.
-const testVariantDefinition = /^var\s+\$elm_explorations\$test\$Test\$Internal\$(ElmTestVariant__\w+|UnitTest|FuzzTest|Labeled|Skipped|Only|Batch)\s*=\s*(?:\w+\(\s*)?function\s*\([\w, ]*\)\s*\{\s*return *\{\s*\$:\s*(['"])\1\2/gm;
+const testVariantDefinition = /^var\s+\$elm_explorations\$test\$Test\$Internal\$(?:ElmTestVariant__\w+|UnitTest|FuzzTest|Labeled|Skipped|Only|Batch)\s*=\s*(?:\w+\(\s*)?function\s*\([\w, ]*\)\s*\{\s*return *\{/gm;
```
harrysarson, line 47: What does `(?:\w+\(\s*)?` do?

harrysarson (approved):

lib/Generate.js
lydell: It matches the `F2` wrapper that the `Labeled` variant has:

```js
var $elm_explorations$test$Test$Internal$Labeled = F2(
	function (a, b) {
		return {$: 'Labeled', a: a, b: b};
	});
```

I tried to future-proof by not caring about the exact name.

harrysarson: ah yes! Of course. cool

PR #474, opened 2020-11-15
Replace Travis CI and Appveyor with GitHub Actions

lydell:
Because of https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

I’m testing the actions over here: https://github.com/lydell/node-test-runner/pull/2


lydell:
I’ve now got it all working and I’m satisfied with the setup.

Here’s what we need to do, as far as I understand:

1. Merge this pull request. This should enable the actions for this repo, and make them show up for all new PRs and PRs that update to latest master.

2. Remove the Appveyor hook.

tests/fixtures/tests/Passing/TrickyMultilines.elm
```diff
@@ -20,6 +20,8 @@ var $elm_explorations$test$Test$Internal$Batch = function (a) {
   return {$: 'Batch', a: a};
 };
 """
+        -- Remove CR on Windows.
+        |> String.replace "\u{000D}" ""
```
harrysarson, line 24: What happens with out this line?

avh4: I would guess this is probably needed when a windows machine checks out this file from git.  If `core.autocrlf` is set to true, then git will insert CRs in the multiline strings.  If that's the cause, then ensuring `core.autocrlf=input` when checking out the project should make this line unnecessary.

tests/fixtures/tests/Passing/TrickyMultilines.elm
```diff
@@ -20,6 +20,8 @@ var $elm_explorations$test$Test$Internal$Batch = function (a) {
   return {$: 'Batch', a: a};
 };
 """
+        -- Remove CR on Windows.
```
harrysarson, line 23:
```suggestion
        -- Remove CR on Windows to match manually escaped newlines in the
        -- testCode expectation.
```

tests/fixtures/tests/Passing/TrickyMultilines.elm
harrysarson: Yeah, because below we check this string against a manually escaped string. I added a bit of extra explanation which hopefully will prevent my future self from becoming confused again.

harrysarson:
I merged this quickly because it is blocking everything else.

PR #464, opened 2020-10-30
Moving to async/await: lib/Runner.js

chebro:
This PR attempts to close #463

Issue #463, opened 2020-10-29
Replace `new Promise` and `.then` with async/await


harrysarson:
We no longer support node 8 so there is nothing holding us back from asyncing and awaiting.

---

To do this

1. Search in files for `new Promise` or `.then`.
2. Replace it with an async function.
3. Make a PR.

It is worth keeping the PR's small (maybe only adding a one or two async/awaits at a time) so we can review them quickly and get them merged!

bhumika7:
Hey! Can you assign this to me?

sravanth-chebrolu:
Hi, I'd like to contribute! Do you recommend any good place to start?

harrysarson:
Hi! Go for it!

sravanth-chebrolu:
I haven't worked with `.then()` blocks in a return statements a lot, it would be awesome if you could help me out here! I'm looking at this bit of code:

```
      return Parser.extractExposedPossiblyTests(filePath).then(
        (possiblyTests) => ({
          moduleName,
          possiblyTests,
        })
      );
```

I don't understand how to return a `Promise.then()`, using `await`, what I could think of is this

```
      possiblyTests = await Parser.extractExposedPossiblyTests(filePath)
      return (possiblyTests) => ({
          moduleName,
          possiblyTests,
        })
```

This will simple wait for the `Promise` to resolve and return the function that follows. But I'm assuming we have to return the `Promise` as it is and let the code that called this function that called it handle the waiting, or am I wrong?

sravanth-chebrolu:
This is the other way I could think of, 
```
      return (possiblyTests = await Parser.extractExposedPossiblyTests(filePath)) => ({
          moduleName,
          possiblyTests,
        })
```

harrysarson:
So the trick it to make the function async so

```js
function xxx(args) {
      ...
      return Parser.extractExposedPossiblyTests(filePath).then(
        (possiblyTests) => ({
          moduleName,
          possiblyTests,
        })
      );
}
```

becomes

```js
async function xxx(args) {
      ...
      const possiblyTests = await Parser.extractExposedPossiblyTests(filePath);
      return {
          moduleName,
          possiblyTests,
      };
}
```

and because it is an async function it will return a promise exactly the same as before.

sravanth-chebrolu:
> So the trick it to make the function async 

[...]

Issue #355, opened 2019-04-14
Error if run from a subdirectory


harrysarson:
### To reproduce this error

In an empty directory run

1. `$ elm init`
1. `$ elm-test init`
2. `$ cd tests`
3. `$ elm-test`

### Output

```
Error reading elm.json: Error: ENOENT: no such file or directory, open '/media/harry/WINDOWS/Users/harry/elm/test-example/tests/elm.json'
```

---
OS: ubuntu
All the elm-test versions 0.19.0 and later have this bug.

lydell:
For anyone how would like to pick this up:

The commands that need `elm.json` assert that an `elm.json` exists in the current working directory:

https://github.com/rtfeldman/node-test-runner/blob/2e2fd090837177f89218e2a4096dce4ea4d91e14/lib/elm-test.js#L380-L389

Those commands should take an elmJson as input instead. Then we could replace the `assertElmJsonInCwd` function with a function that finds the first `elm.json` upwards in the folder hierarchy (passing it on to the commands that need it), or errors saying that no `elm.json` could be found.

Issue #225, opened 2017-11-19
Watcher: Cancel previous compilation and test runs


andys8:
I opened #194 in the past and we took some steps in the right direction but I think the main issue is still existing: `node-test-runner` spawn's processes to compile elm sources, but doesn't cancel them if a new file change triggers a new run.

#### Demo
![elm-test](https://user-images.githubusercontent.com/13085980/32992428-e2d4723e-cd4a-11e7-84da-06c22f67719e.gif)
This shows that touching a file frequently will lead to problems if the timespan is shorter than a complete test run including the compile step. In the example the test run itself isn't long, but you can see the overlapping runs and the cpu utilization. The issue get's worse for projects with a larger code base.

This issue happens in my work flow and I tend to disable the file watcher because its getting unusable. Instead I try to optimize and watch a single test file or start a test run manually.

I see there has been work done in the latest commits regarding the architecture where the supervisor and the workers communicate with each other. I'm not to sure what's going on in detail, so I'd like to discuss possible solutions to improve this issue. In my opinion it could work out to `kill` any spawned process if a new test run is triggered in watch mode. There is one part where the internal logic has to be able to be aborted and there is another that external processes (namely the elm compiler) have to be killed. I previously [mentioned](https://github.com/rtfeldman/node-test-runner/issues/194#issuecomment-319889810) that it would be necessary to be able to stop the spawned process  and migrate to `cross-spawn-with-kill` or hope something is happening here https://github.com/IndigoUnited/node-cross-spawn/pull/44 or use a different solution to control spawned processes (https://nodejs.org/api/child_process.html).
My guess is also that the `node-elm-compiler` has to be adapted, to be cancelable. it contains also a dependency to `cross-spawn@4.0.0`.

I'm happy to discuss if this solution to cancel spawned processes makes any sense at all. If there could be problems aborting the `elm-compiler` and having an inconsistent state. Perhaps there are other options and debouncing/thottling the actual test runs will solve the issue.

PR #475, opened 2020-11-21
Improve watcher

lydell:
This makes the watcher faster and more stable! As a bonus, this made the code much more organized.

- Concentrate all `process.exit` in elm-test.js (except one in Install.js where I found no clean solution). Besides making the code easier to follow, it allows the watcher to continue running (and recover) even if there are errors.

- Don’t crash/stop running tests if files are removed.

- Pick up new Elm files added.

- Fixes https://github.com/rtfeldman/node-test-runner/issues/225

  We already had code for waiting for the previous run to finish before starting the next run, but it queued up every single file change/add/remove event and ran that many runs one after the other. Now, for all watch events during a run we run _once_ after the current run for all of them. We also sleep 100 ms at the start of each run to catch more events happening close by (such as when adding/removing a whole folder of tests). (Note: We used to sleep at least 500 ms via the [awaitWriteFinish](https://github.com/paulmillr/chokidar#performance) chokidar option.)

  I’ve also tested manually that running elm-format on save doesn’t cause double test runs: https://github.com/rtfeldman/node-test-runner/issues/194

- Fixes https://github.com/rtfeldman/node-test-runner/issues/355

  We now find the closest elm.json upwards in the folder hierarchy.

- Resolves https://github.com/rtfeldman/node-test-runner/pull/453#discussion_r506275471

  The watcher now listens for changes to elm.json and properly recalculates things. I thought that I would have to recalculate some things only when elm.json changes for performance, but after making the globbing faster there was no need so we simply rebuild everything on each run.

- Closes https://github.com/rtfeldman/node-test-runner/issues/463 and closes https://github.com/rtfeldman/node-test-runner/pull/464

  I can’t find any more places to use async/await.

- elm.json is now decoded properly in one place rather than here and there and just being `any`. We now have 0 explicit `any` type annotations!

- elm-test.js now only contains CLI logic and subcommand dispatching. The code for running tests have been moved to RunTests.js and to FindTests.js (which used to be called Runner.js).

- Improved .flowconfig that is stricter. Closes https://github.com/rtfeldman/node-test-runner/issues/478

- And lots of little code improvements along the way! I only touched things that I needed to change anyway in order to not make the diff balloon too much.

I’ve done some manual testing on Windows, Mac and Linux to make sure things work like they should.


pipefail:
Just want to say this looks great! Awesome work and addresses a lot of important issues.

lib/ElmJson.js
```diff
@@ -0,0 +1,193 @@
...
+  if (json == null || typeof json !== 'object' || Array.isArray(json)) {
+    throw new Error(
+      `Expected ${what} to be an object, but got: ${stringify(json)}`
+    );
+  }
+  return json;
+}
+
+function stringify(json /*: mixed */) /*: string */ {
+  const maybeString = JSON.stringify(json);
```
pipefail, line 182: Minor comment: `JSON.stringify` can throw a `TypeError`

lib/ElmJson.js
```diff
@@ -0,0 +1,193 @@
...
+  } catch (error) {
+    throw new Error(
+      `${elmJsonPath}\nFailed to read elm.json:\n${error.message}`
+    );
+  }
+}
+
+function readHelper(elmJsonPath /*: string */) /*: typeof ElmJson */ {
+  const json = parseObject(
+    JSON.parse(fs.readFileSync(elmJsonPath, 'utf8')),
```
pipefail, line 64: Minor comment: `JSON.parse` can throw a `SyntaxError`

lib/Parser.js
```diff
@@ -312,16 +312,19 @@ void OnParserTokenResult;
...
   return {
     tag: 'ParseError',
-    message: `Expected ${expectedDescription} but got: ${JSON.stringify(
-      actual
-    )}`,
+    message: `Expected ${expectedDescription} but got: ${stringify(actual)}`,
   };
 }
 
+function stringify(json /*: mixed */) /*: string */ {
```
pipefail, line 323: Duplicated func definition?

lib/RunTests.js
```diff
@@ -0,0 +1,274 @@
...
+) /*: Promise<number> */ {
+  let watcher = undefined;
+  let watchedPaths /*: Array<string> */ = [];
+  let runsExecuted /*: number */ = 0;
+  let currentRun /*: Promise<void> | void */ = undefined;
+  let queue /*: typeof Queue */ = [];
+
+  async function run() /*: Promise<number> */ {
+    try {
+      // Don’t delay the first run (that’s the only time the queue is empty).
```
pipefail, line 109: Sensible optimization

lib/RunTests.js
```diff
@@ -0,0 +1,274 @@
...
+    infoLog(report, 'Running in watch mode');
+
+    const onRunFinish = () => {
+      if (queue.length > 0) {
+        clearConsole(report);
+        infoLog(report, watcherEventMessage(queue));
+        currentRun = run().then(onRunFinish);
+      } else {
+        infoLog(report, chalk.blue('Watching for changes...'));
+        currentRun = undefined;
```
pipefail, line 216: minor: if not planning to take advantage of invoking default arg handling semantics passing `currentRun` anywhere, then recommend avoiding `undefined` assignment as it is somewhat uniquely quirky to JS. i.e. prefer `null`

pipefail (commented):
Looks great! A few minor comments but nothing serious. Looking forward to this PR, been dealing with several of these issues for a while now.

pipefail (approved):

lib/ElmJson.js
lydell: Yes, I have tests for that. Everything in this file throws errors when something’s invalid.

lib/ElmJson.js
lydell: True! However, not in this particular file. The only things I’m ever stringifying are things that just came from a `JSON.parse`.

lib/Parser.js
lydell: Yes! I’m not sure if it makes sense to have a shared definition.

lib/RunTests.js
lydell: I disagree – I find `null` much quirkier than `undefined`

pipefail: Heh, ok

lib/ElmJson.js
harrysarson: Maybe a comment at the top of the file noting that all these functions may throw would help here.

lib/FindTests.js
```diff
@@ -0,0 +1,263 @@
...
+const gracefulFs = require('graceful-fs');
+const fs = require('fs');
+const glob = require('glob');
+const path = require('path');
+const Parser = require('./Parser');
+const Project = require('./Project');
+
+void Project;
+
+function flatMap/*:: <T, U> */(
```
harrysarson, line 12:
```suggestion
// Remove in favour of `Array.flatMap` once node 10 is eol.
function flatMap/*:: <T, U> */(
```

lib/FindTests.js
```diff
@@ -0,0 +1,263 @@
...
+}
+
+function resolveCliArgGlob(
+  fileGlob /*: string */,
+  projectRootDir /*: string */
+) /*: Array<string> */ {
+  return flatMap(
+    // Globs passed as CLI arguments are relative to CWD, while elm-test
+    // operates from the project root dir.
+    glob.sync(path.relative(projectRootDir, path.resolve(fileGlob)), {
```
harrysarson, line 65:
```suggestion
    const globRelativeToProjectRoot = path.relative(projectRootDir, path.resolve(fileGlob);
    glob.sync(globRelativeToProjectRoot), {
```

Just makes it clearer that the comment refers the the glob we pass to `glob.sync` rather than the `glob.sync` call in general.

lib/FindTests.js
```diff
@@ -0,0 +1,263 @@
+// @flow
```
harrysarson, line 1: Are there any changes to these functions, or have they just been moved?

lib/ElmJson.js
lydell: Sure, I can add a comment, but isn’t this kind of obvious? The file is full of `throw new Error` and it feels like well-known stuff that reading files and parsing JSON throws errors if they fail?

lib/FindTests.js
```diff
@@ -0,0 +1,269 @@
...
+        // We’re reading files asynchronously in a loop here, so it makes sense
+        // to use graceful-fs to avoid “too many open files” errors.
+        gracefulFs.createReadStream
+      ).then((possiblyTests) => ({
+        moduleName,
+        possiblyTests,
+      }));
+    })
+  );
+}
```
lydell, line 163: Moved from Runner.js (really, Runner.js was renamed to FindTests.js because that was all it was doing). The only change is that the function takes a `project: Project` parameter instead of `sourceDirs: Array<string>, isPackageProject: boolean`.

lib/FindTests.js
```diff
@@ -0,0 +1,269 @@
...
+
+…matches no source directory! Imports won't work then.
+
+${
+  isPackageProject
+    ? 'Move it to tests/ or src/ in your project root.'
+    : 'Move it to tests/ in your project root, or make sure it is covered by "source-directories" in your elm.json.'
+}
+  `.trim();
+}
```
lydell, line 179: From Runner.js. No changes.

lib/FindTests.js
```diff
@@ -0,0 +1,269 @@
...
+
+…matches more than one source directory:
+
+${matchingSourceDirs.join('\n')}
+
+Edit "source-directories" in your elm.json and try to make it so no source directory contains another source directory!
+
+${note}
+  `.trim();
+}
```
lydell, line 203: From Runner.js. The only change is adding the note about the tests/ folder counting as a source directory too. That happens when the user has `"."` in their source directories, which elm-test does not support if you want to use the tests/ directory. See also: https://discourse.elm-lang.org/t/elm-test-error-with-directory-name-structure/6552

lib/FindTests.js
```diff
@@ -0,0 +1,269 @@
...
+${moduleName}
+
+…but module names need to look like for example:
+
+Main
+Http.Helpers
+
+Make sure that all parts start with an uppercase letter and don't contain any spaces or anything like that.
+  `.trim();
+}
```
lydell, line 226: From Runner.js. No changes.

lib/FindTests.js
```diff
@@ -0,0 +1,269 @@
...
+try calling elm-test with a glob such as: elm-test "src/**/*Tests.elm"
+      `.trim()
+    : `
+No files found matching:
+
+${testFileGlobs.join('\n')}
+
+Are the above patterns correct? Maybe try running elm-test with no arguments?
+      `.trim();
+}
```
lydell, line 248: From elm-test.js. The only change is the first line of the error message – it used to always be “No .elm files found in the tests/ directory.”. Now, it depends on the `noFilesFoundInTestsDir` function – see the next review comment.

lib/FindTests.js
```diff
@@ -0,0 +1,269 @@
...
+    const stats = fs.statSync(testsDir);
+    return stats.isDirectory()
+      ? 'No .elm files found in the tests/ directory.'
+      : `Expected a directory but found something else at: ${testsDir}\nCheck it out! Could you remove it?`;
+  } catch (error) {
+    return error.code === 'ENOENT'
+      ? 'The tests/ directory does not exist.'
+      : `Failed to read the tests/ directory: ${error.message}`;
+  }
+}
```
lydell, line 262: This is new. It replaces dead code in Generate.js:

```js
  if (needToCareAboutTestsDir) {
    if (!fs.existsSync(testRootDir)) {
      console.error(
        'Error: ' +
          testRootDir +
          ' does not exist. Please create a tests/ directory in your project root!'
      );
      process.exit(1);
    }

    if (!fs.lstatSync(testRootDir).isDirectory()) {
      console.error(
        'Error: ' +
          testRootDir +
          ' exists, but it is not a directory. Please create a tests/ directory in your project root!'
      );
      process.exit(1);
    }
  }
```

lib/Generate.js
```diff
@@ -59,110 +61,51 @@ function addKernelTestChecking(content) {
...
-          ' exists, but it is not a directory. Please create a tests/ directory in your project root!'
-      );
-      process.exit(1);
-    }
-  }
+function getGeneratedSrcDir(generatedCodeDir /*: string */) /*: string */ {
+  return path.join(generatedCodeDir, 'src');
+}
+
+function generateElmJson(project /*: typeof Project.Project */) /*: void */ {
```
lydell, line 68: This function was doing a lot of calculations that other parts of the code needs as well, so I moved those things into Project.js. Now, this function instead just takes a `project: Project` and could be simplified.

lib/Generate.js
```diff
@@ -59,110 +61,51 @@ function addKernelTestChecking(content) {
...
-  const needToCareAboutTestsDir =
-    hasBeenGivenCustomGlobs === false || isPackageProject === true;
-
-  // we add the tests dir as a source if:
-  // - we decided above we need to care about it
-  // - if it exists on disk: this supports the case where we have all our tests in tests/ but
-  //   want to pass a glob to only run one of the test files
-  const shouldAddTestsDirAsSource =
-    needToCareAboutTestsDir ||
-    fs.existsSync(path.join(projectRootDir, 'tests'));
```
lydell, line 86: This has been moved to Project.js, and has been greatly simplified. All we need to do is – if tests/ exists, then add it as a source directory, otherwise don’t, regardless of whether testing a package or an application.

lib/Generate.js
```diff
@@ -59,110 +61,51 @@ function addKernelTestChecking(content) {
...
-
-    if (!fs.lstatSync(testRootDir).isDirectory()) {
-      console.error(
-        'Error: ' +
-          testRootDir +
-          ' exists, but it is not a directory. Please create a tests/ directory in your project root!'
-      );
-      process.exit(1);
-    }
-  }
```
lydell, line 106: As mentioned earlier, this code was dead due to earlier checks. I moved these messages into another error message.

lib/Install.js
```diff
@@ -1,86 +1,71 @@
...
-    (isPackage && elmJson['test-dependencies'].hasOwnProperty(packageName)) ||
-    (!isPackage &&
-      elmJson['test-dependencies']['direct'].hasOwnProperty(packageName))
+    elmJson.type === 'package'
+      ? elmJson['test-dependencies'].hasOwnProperty(packageName)
+      : elmJson['test-dependencies'].direct.hasOwnProperty(packageName)
   ) {
-    console.log('It is already installed!');
-    return;
+    return 'AlreadyInstalled';
```
lydell, line 42: Returning instead of logging here is nice because when in this scenario:

1. `elm-test init`
2. Delete tests/
3. `elm-test init` again, to regenerate tests/

Now, “It is already installed!” won’t be logged in that case. That message was only ever meant for `elm-test install user/package` when `user/package` was already installed.

lib/Install.js
```diff
@@ -1,86 +1,71 @@
...
+
+  ElmJson.write(generatedCodeDir, tmpElmJson);
+
+  const result = spawn.sync(pathToElmBinary, ['install', packageName], {
+    stdio: 'inherit',
+    cwd: generatedCodeDir,
+  });
+
+  if (result.status !== 0) {
+    process.exit(result.status);
```
lydell, line 63: I tried to concentrate all `process.exit` in elm-test.js, but I couldn’t figure out a good way to do it here. In this case we have run `elm install` and printed its output to the console. If `elm install` failed we just want to exit with `elm install`’s error output and not print anything else or so. I think this is fine.

lib/Parser.js
```diff
@@ -312,16 +312,19 @@ void OnParserTokenResult;
 
 function expected(
   expectedDescription /*: string */,
-  actual /*: any */
+  actual /*: mixed */
```
lydell, line 315: This was the last `any` annotation left after making ElmJson.js (we used to have a bunch of `elmJson: any` parameters to functions). So I got rid of this one as well. That required the `stringify` function to please Flow, since unexpectedly `JSON.stringify(undefined) === undefined` (all other inputs than `undefined` produce a string or throw an error).

lib/RunTests.js
```diff
@@ -0,0 +1,274 @@
...
+  }
+}
+
+function clearConsole(report /*: typeof Report.Report */) {
+  if (report === 'console') {
+    process.stdout.write(
+      process.platform === 'win32' ? '\x1B[2J\x1B[0f' : '\x1B[2J\x1B[3J\x1B[H'
+    );
+  }
+}
```
lydell, line 47: From elm-test.js. No changes.

lib/RunTests.js
```diff
@@ -0,0 +1,274 @@
...
+  if (filePaths.length === 1) {
+    const { event, filePath } = queue[0];
+    return `${filePath} ${event}${suffix}`;
+  }
+
+  const events = Array.from(new Set(queue.map(({ event }) => event))).sort();
+  return `${filePaths.length} files ${events.join('/')}${suffix}`;
+}
+
+function runTests(
```
lydell, line 84: This function is the `test` function plus the `runTests` function from elm-test.js. I started by merging those two into one – they were a very arbitrary split, where the last thing `test` did was calling `runTests`. I find it much easier to understand when they are one.

Then there is some new code for the improved queuing of tests and how the watcher is set up.

lib/RunTests.js
```diff
@@ -0,0 +1,274 @@
...
+    report: typeof Report.Report,
+    seed: number,
+    fuzz: number,
+  } */
+) /*: Promise<number> */ {
+  let watcher = undefined;
+  let watchedPaths /*: Array<string> */ = [];
+  let runsExecuted /*: number */ = 0;
+  let currentRun /*: Promise<void> | void */ = undefined;
+  let queue /*: typeof Queue */ = [];
```
lydell, line 105: I really like how all mutable state ended up in one place.

lib/Supervisor.js
```diff
@@ -6,7 +6,6 @@ const fs = require('fs');
 const net = require('net');
 const split = require('split');
 const XmlBuilder = require('xmlbuilder');
-// $FlowFixMe: Flow marks this line as an error (and only in this file!) but then it knows the types of `Report` anyway. This is the “error”: Cannot build a typed interface for this module. You should annotate the exports of this module with types. Missing type annotation at declaration of variable `Report`: [signature-verification-failure]
```
lydell, line 9: This stopped being a problem when I improved .flowconfig. I have no idea why, but I’m not complaining.

lib/FindTests.js
lydell: A lot of the diff is actually just moved stuff! I’ve added comments to show what is just moved and what has changed.

lib/ElmJson.js
harrysarson: Do we not usually use a `Result` type now in the test runner? If someone saw our use of result types in some files they might expect us to handle all errors with a  `Result` type. I was just thinking it might help to clarify which paradigm we use in which file?

lydell: I got rid of `Result` because it was much easier to deal with regular exceptions in the end. Especially in async code.

lib/Project.js
```diff
@@ -0,0 +1,117 @@
...
+  generatedCodeDir: string,
+  testsSourceDirs: Array<string>,
+  elmJson: typeof ElmJson.ElmJson,
+} */ = {
+  rootDir: '',
+  testsDir: '',
+  generatedCodeDir: '',
+  testsSourceDirs: [],
+  elmJson: ElmJson.ElmJson,
+};
```
harrysarson, line 21: Thoughts on making this a class with a constructor?

lydell: Yes, one – why? 😄

lib/Supervisor.js
harrysarson: nice, I won't complain either :)

tests/fixtures/write-elm-json/elm.input.json
```diff
@@ -0,0 +1,25 @@
...
+    "indirect": {
+      "elm/html": "1.0.0",
+      "elm/virtual-dom": "1.0.2"
+    }
+  },
+  "nonStandardFieldEnd": {
+    "a": 1,
+    "b": 2
+  }
+}
```
harrysarson, line 25:
```suggestion
}

```

lydell: What is the thinking here?

harrysarson:
Thanks for the comments explaining the rearrangements; they are really helpful. I am getting there:

image

---

Thanks @pipefail too for your help, much appreciated :)

lib/Generate.js
```diff
@@ -190,39 +144,24 @@ function generateMainModule(
...
     makeOptsCode(fuzz, seed, report, testFileGlobs, testFilePaths, processes)
   );
 
-  // Generate a filename that incorporates the hash of file contents.
-  // This way, if you run e.g. `elm-test Foo.elm` and then `elm-test Bar.elm`
-  // and then re-run `elm-test Foo.elm` we still have a cached `Main` for
-  // `Foo.elm` (assuming none of its necessary imports have changed - and
-  // why would they?) so we don't have to recompile it.
-  const salt = Murmur.murmur3(testFileBody);
-  const moduleName = 'Main' + salt;
```
lydell, line 207: The comment here says that the use case is:

```
elm-test Foo.elm
elm-test Bar.elm
# We should have a ready-to-go generated MainXXXX.elm file here:
elm-test Foo.elm
```

But that’s not true. There will be a unique MainXXXX.elm file for every invocation there. Why? Because the murmur hash is based on the MainXXXX.elm file contents (by design, according to the comment), and the file contents contain the _seed,_ which is randomly generated on startup. So the file contents _won’t_ be exactly the same on the second `elm-test Foo.elm` run, so it’ll get a new hash and a new generated file name and as such not benefit from any “cached” stuff.

What about watch mode, then? Was there any benefit from the hashing there? Because there you run with the same seed all the time, so the file contents shouldn’t change that often? That is all true – but removing the hash makes no difference. Before, you’d have Main3541244.elm for a bunch of runs, now you have just Main.elm for a bunch of runs.

The only time we _could_ reuse an old MainXXXX.elm is when you rerun tests with the `--seed` flag. But even then I don’t see the point – we still do all the compilation just as usual without skipping anything, and a comment below even mentions the importance of never skipping writing the generated .elm file to update the timestamp:

> // Always write the file, in order to update its timestamp. This is important,
> // because if we run `elm-make Main123456.elm` and that file's timestamp did
> // not change, elm-make will short-circuit and not recompile *anything* - even
> // if some of Main's dependencies (such as an individual test file) changed.

So I feel confident the hashing made no difference. I’ve not noticed any slowdown or anything from removing it. On the contrary, not hashing should be faster since we now do less work! And we get rid of one more dependency.

Another benefit is that `elm-stuff/` no longer grows indefinitely, with one new `MainXXXX.elm` file – along with .elmi and .elmo files – per `elm-test` invocation, basically. We now overwrite the same `Main.elm` all the time.

lib/Generate.js
```diff
@@ -59,110 +60,51 @@ function addKernelTestChecking(content) {
...
+  return path.join(generatedCodeDir, 'src');
+}
+
+function generateElmJson(project /*: typeof Project.Project */) /*: void */ {
+  const generatedSrc = getGeneratedSrcDir(project.generatedCodeDir);
 
-  fs.mkdirSync(generatedCodeDir, { recursive: true });
   fs.mkdirSync(generatedSrc, { recursive: true });
 
   let testElmJson = {
```
harrysarson, line 72:
```suggestion
  // TODO(harrysarson): use object.assign here instead of mutable objects.
  let testElmJson = {
```

lib/Generate.js
harrysarson: Brilliant!

lib/Generate.js
```diff
@@ -190,39 +144,24 @@ function generateMainModule(
...
-  fs.mkdirSync(mainPath, { recursive: true });
+  // Make sure the generated code dir exists so we can write the file there.
+  fs.mkdirSync(path.dirname(mainModule.path), { recursive: true });
 
   // Always write the file, in order to update its timestamp. This is important,
-  // because if we run `elm-make Main123456.elm` and that file's timestamp did
-  // not change, elm-make will short-circuit and not recompile *anything* - even
+  // because if we run `elm make Main.elm` and that file's timestamp did
+  // not change, `elm make` will short-circuit and not recompile *anything* - even
   // if some of Main's dependencies (such as an individual test file) changed.
```
harrysarson, line 163: Is this still true? I think the elm compiler is now smarter than this.

lib/Install.js
```diff
@@ -1,86 +1,71 @@
...
-  const dirPath = path.join(
-    Compiler.getGeneratedCodeDir(process.cwd()),
-    'install'
-  );
+function install(
+  project /*: typeof Project.Project */,
+  pathToElmBinary /*: string */,
+  packageName /*: string */
+) /*: 'SuccessfullyInstalled' | 'AlreadyInstalled' */ {
+  const generatedCodeDir = path.join(project.generatedCodeDir, 'install');
```
harrysarson, line 17:
```suggestion
  const installationScratchDir = path.join(project.generatedCodeDir, 'install');
```

Just because the phrase `generatedCodeDir` already means the parent dir of this dir so using a different name (ignore my suggestion if you have a better idea) helps disambiguate them.

lib/Install.js
harrysarson: Because `init` calls `install` to add the test package as a dev dep. Yep :+1:

lib/Install.js
harrysarson: :+1:

lib/ElmJson.js
harrysarson: Makes sense :+1: both results and exceptions have their benefits but consistency is most important

lib/elm-test.js
```diff
@@ -354,16 +45,89 @@ const parsePositiveInteger = (flag /*: string */) => (
...
+    return Report.parse(string);
+  } catch (error) {
+    console.error(`error: option '${flag}' ${error.message}`);
+    throw process.exit(1);
   }
 };
 
+function findClosest(
+  dir /*: string */,
+  dirToPath /*: (dir: string) => string */
```
harrysarson, line 58: We only call this function in one place so there is only ever one `dirToPath` value (which is `ElmJson.getPath`) could we hard code it? I think it would make the purpose of this function clearer and the logic more straightforward.

lib/elm-test.js
```diff
@@ -354,16 +45,89 @@ const parsePositiveInteger = (flag /*: string */) => (
...
+(If elm is not on your path or is called something different the --compiler flag might help.)`
+        : `The elm executable passed to --compiler must exist and be exectuble. Got: ${compiler}`
+    );
+  }
+}
+
+// Unfortunately commander is very permissive about extra arguments. Therefore,
+// we manually check for excessive arguments.
+// See: https://github.com/tj/commander.js/issues/1268
+function handleTooManyArgs(action) {
```
harrysarson, line 107: nice, this and getPathToElmBinary not being top level makes everything so much nicer

lib/elm-test.js
```diff
@@ -427,106 +178,94 @@ function main() {
...
+      );
     });
 
   program
-    // Hack: This command is named `tests` so that `elm-test tests` means the same
-    // as `elm-test tests/` (due defaulting to `tests/` if no arguments given).
-    .command('tests [globs...]', { hidden: true, isDefault: true })
+    // Hack: This command has a name that isn’t likely to exist as a directory.
+    // If the command where called for example “tests” then `elm-test tests src`
+    // would only run tests in `src/`, not `tests/`.
```
harrysarson, line 252:
```suggestion
    // Hack: This command has a name that isn’t likely to exist as a directory
    // containing tests. If the command were instead called "tests" then
    // commander would interpret the `"tests"` in `elm-test tests src` as a
    // command and only run tests in `src/`, ignoring all files in `tests/`.
```

Feel free to ignore

lib/Parser.js
harrysarson: yeah not too worried about this duplication, is pretty simple function.

lib/Generate.js
harrysarson: sounds good!

lib/Parser.js
harrysarson: nice, is there a lint to prevent use of any?

lib/Project.js
harrysarson: I will go and find another hill to die on and leave you in peace.

tests/fixtures/write-elm-json/elm.input.json
harrysarson: I like files to have trailing new lines?

harrysarson:
Just got runTests to go. Nearly there :tada:

lib/Generate.js
lydell: You seem to be right! I can’t reproduce this problem.

I’ve also tested watch mode a bit at my work code base, and always generating the same file name (Main.elm) seems to cause no issues.

I removed the comment.

lib/Generate.js
lydell: It was easy to change the code to not use mutation here so I did the change instead of committing this comment.

lib/Install.js
lydell: Fixed!

lib/Parser.js
lydell: There’s [eslint-plugin-flowtype/no-weak-types](https://github.com/gajus/eslint-plugin-flowtype#no-weak-types). I could install that plugin and enable some rules in another PR – I think this one has enough changes already.

tests/fixtures/write-elm-json/elm.input.json
lydell: Ah, of course! I misinterpreted the suggested change as wanting a blank line at the end of the file, but you’re right the ending newline is missing here. I’ve added one now.

lib/elm-test.js
lydell: 👍 Fixed!

lib/elm-test.js
lydell: Good idea, I made the change.

lib/Generate.js
harrysarson: Some of the test runner seems really old. Is so nice to get rid of this tech debt.

lib/Generate.js
harrysarson: nice!

tests/fixtures/write-elm-json/elm.input.json
harrysarson: yeah the github diff UI leaves room for improvement here.

harrysarson (approved):
Happy!

harrysarson:
I like the queuing a lot and thanks for the help finding which code had moved where. You ready for me to merge?

lydell:
I’m ready! :shipit:

Issue #484, opened 2021-01-12
Add support for darwin-arm64


czepluch:
I just got my hands on an Apple Silicon based macbook, and after installing nvm and compiling the newest version of node I get the following error when trying to install `elm-test`:

```
Error: No binaries are available for your platform: darwin-arm64
```

lydell:
Hi! I think this won’t happen on the master branch. Can you try it? Something like this:

```
npm install --save-dev rtfeldman/node-test-runner
npx elm-test
```

czepluch:
Still getting the same error when trying above. Maybe I can just install an older non-arm version of npm and then run the Rosetta translated version.

harrysarson:
Can you install elm-json?

lydell:
Ah, shoot. That’s because when you install from GitHub, npm also install all _dev_ dependencies. And in  our dev dependencies we have elm-review, which eventually depends on the binwrap package which is where “No binaries are available for your platform” message comes from.

A new version of elm-review was just released though which no longer depends on binwrap, so if we update to that version I think `npm install --save-dev rtfeldman/node-test-runner` will start working.

lydell:
I made a pull request: https://github.com/rtfeldman/node-test-runner/pull/485

@czepluch Can you try installing from the PR?

```
npm i --save-dev lydell/node-test-runner#update-elm-review
```

czepluch:
> Can you install elm-json?

Yes.

> npm i --save-dev lydell/node-test-runner#update-elm-review

Worked!

harrysarson:
Now fixed on master. Thanks for the report :)

czepluch:
Thanks for the quick response :D

PR #485, opened 2021-01-13
Update elm-review

lydell:
The latest version of elm-review uses elm-tooling to install elm-json, so it no longer depends on binwrap.

When installing packages from GitHub using npm, npm also install all dev dependencies (!). So with this PR people can do `npm install --save-dev rtfeldman/node-test-runner` to try the latest master without getting binwrap. That’s useful for example for https://github.com/rtfeldman/node-test-runner/issues/484

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -127,9 +127,9 @@
       }
     },
     "@types/node": {
-      "version": "14.11.8",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-14.11.8.tgz",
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -227,99 +227,28 @@
...
-      "integrity": "sha512-Y+Wvypk3JhH5GPZAvlwJAWOVH/OsOhQMSj37vySuWHwQivoALplPxfBA8b973rFJI7OS+O+1YmmYXIiEXVMAcw==",
-      "dev": true,
-      "requires": {
-        "mustache": "^3.0.1",
-        "request": "^2.88.0",
-        "request-promise": "^4.2.4",
-        "tar": "^4.4.10",
-        "unzip-stream": "^0.3.0"
-      }
-    },
```
lydell, line 316: Yay!

harrysarson:
Thanks!

PR #486, opened 2021-01-15
Upgrade to commander 7

lydell:
This allows getting rid of all workarounds we had for commander.

### DIFF
--- a/lib/elm-test.js
+++ b/lib/elm-test.js
@@ -1,6 +1,6 @@
 // @flow
 
-const { program } = require('commander');
+const { InvalidOptionArgumentError, Option, program } = require('commander');
 const fs = require('fs');
 const os = require('os');
 const path = require('path');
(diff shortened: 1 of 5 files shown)


lib/elm-test.js
```diff
@@ -150,23 +98,24 @@ function main() {
...
     )
-    .option(
-      '--seed <int>',
-      'Run with a previous fuzzer seed',
-      parsePositiveInteger('--seed <int>'),
-      Math.floor(Math.random() * 407199254740991) + 1000
+    .addOption(
+      new Option('--seed <int>', 'Run with a previous fuzzer seed')
+        .default(Math.floor(Math.random() * 407199254740991) + 1000, 'random')
+        .argParser(parsePositiveInteger)
```
harrysarson, line 104: Out of interest what is going here? The `'random'` looks a bit special.

lib/elm-test.js
```diff
@@ -234,7 +179,10 @@ function main() {
       const make = async () => {
         await Generate.generateElmJson(project, () => null);
         await Compile.compileSources(
-          FindTests.resolveGlobs(testFileGlobs, project.rootDir),
+          FindTests.resolveGlobs(
+            testFileGlobs.length === 0 ? [project.testsDir] : testFileGlobs,
+            project.rootDir
+          ),
```
harrysarson, line 185: Could we add a unit test?

harrysarson (approved):
Looks good (with a regression test)! Thanks :)

lib/elm-test.js
lydell: Now, the helps looks like:

```
  --seed <int>       Run with a previous fuzzer seed (default: random)
```

Previously, it was:

```
  --seed <int>                   Run with a previous fuzzer seed (default: 390695695167635)
```

With a different number each time.

So it wasn’t clear if the default seed was a constant or what it was.

lib/elm-test.js
lydell: Added! :+1:

lib/elm-test.js
harrysarson: Nice!

harrysarson:
Does the elm make test fail without the fix? And does the readme need updating? Otherwise we are good to merge :)

lydell:
> Does the elm make test fail without the fix?

Yes, I tried it without the fix and they failed with the expected error messages.

> does the readme need updating?

I don’t think so. Was there anything in particular you were thinking about? Either way, I’ll update the readme PR (#473) after this is merged to fix conflicts and see if there’s any more syncing to do.

harrysarson:
I was thinking that we embed the help output in the readme. However, that is probably only true in #473.

PR #491, opened 2021-01-30
Fix --no-color

lydell:
We used to implicitly support `elm-test --no-color` via [chalk.supportsColor](https://github.com/chalk/chalk#chalksupportscolor).

However, when we switched to commander, unknown flags are now errors. Luckily, `FORCE_COLOR=0` still works.

This PR adds `--no-color` and `--color`, so they aren’t treated as unknown flag errors and so they appear in `--help`. I also added `FORCE_COLOR` to `--help`.

Note: `--color=256` is not supported, only `--color`. This is because we only use basic colors, and we don’t want `tests/Foo.elm` to be slurped as the argument to `--color` in `elm-test --color tests/Foo.elm`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,8 +157,6 @@ elm-test --report json
 
 ### --no-color
 
-> **Note:** `--no-color` and `--color` are broken in version 0.19.1-revision5. See [#491](https://github.com/rtfeldman/node-test-runner/pull/491).
-
 Disable colored console output.
 
 Colors are also disabled when you pipe the output of `elm-test` to another program. You can use `--color` to force the colors back.
--- a/README.md
+++ b/README.md
@@ -157,6 +157,8 @@ elm-test --report json
 
 ### --no-color
 
+> **Note:** `--no-color` and `--color` are broken in version 0.19.1-revision5. See [#491](https://github.com/rtfeldman/node-test-runner/pull/491).
+
 Disable colored console output.
 
 Colors are also disabled when you pipe the output of `elm-test` to another program. You can use `--color` to force the colors back.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -155,6 +155,18 @@ Specify which format to use for reporting test results. Valid options are:
 elm-test --report json
 ```
 
+### --no-color
+
+> **Note:** `--no-color` and `--color` are broken in version 0.19.1-revision5. See [#491](https://github.com/rtfeldman/node-test-runner/pull/491).
```
harrysarson, line 160: How would you feel about adding a changelog entry and moving this comment there. I worry that if we add this to the readme it will stay there for ages long after we have forgotten about the revision 5 release?

harrysarson: Especially if we make a revison6 straight away.

lydell: I added it here because I thought it would take a long time to get a new release out. I’ll move it.

harrysarson:
Happy to not block merging on this but this PR only fixes the colouring that the test-runner does directly. The elm compiler does not respect `FORCE_COLOR` or have a reliable `--no-color` flag (nothing we can do about that) but that means that compiler errors appear with colors even when the `--no-color` flag is given to elm-test. 

There is some ansi color codes we make in elm ([here](https://github.com/rtfeldman/node-test-runner/blob/728c13af4aa0d6b485ae91226246755dc65100c5/elm/src/Console/Text.elm)) but we check `require('chalk').supportsColor` before using them so these ansi color codes will not be inserted when `--no-color` is passed. Nice.

lydell:
I can take a quick look to see if it’s easy to disable colors for the compiler output as well.

Issue #495, opened 2021-02-12
Using stdout or stderr for the error message when no test was found?


mpizenberg:
Currently, the console report writes to stderr, the json report writes an equivalent escaped string to stdout, and the junit report crashes with the following message, repeated 8 times (I guess since I have 8 threads):
```
Error: connect ECONNREFUSED /tmp/elm_test-266283.sock
    at PipeConnectWrap.afterConnect [as oncomplete] (net.js:1146:16) {
  errno: -111,
  code: 'ECONNREFUSED',
  syscall: 'connect',
  address: '/tmp/elm_test-266283.sock'
}
```
Except for the junit crash which is probably a bug, it could make sense to have different behavior. Console output reports all failure logs in stderr, and json output uses an escaped string written to stdout probably because that's what tools are reading. I just wanted to point at the differences and maybe get some insight on those decisions since I have to choose which output to use for elm-test-rs.

lydell:
I looked into this and it’s a regression since 0.19.1-revision5. Before that, a human readable message was written to stderr regardless of report format. I’ll make a PR to restore that.

mpizenberg:
Alright I'll add an error message to stderr in all cases too!

PR #496, opened 2021-02-13
Fix "no tests found" errors for json and junit

lydell:
Fixes https://github.com/rtfeldman/node-test-runner/issues/495

I consider this something that _should_ be a compile time error. It isn’t currently because we run the JS to figure out which exposed values are of type Test. But it used to be in 0.19.1-revision4 and older.

All compile time errors are printed human readable to stderr. This PR makes sure that happens for this error as well. There was already code in place for this, but it was only for the `console` reporter. I changed it to apply to all reporters.

I have not been able to figure out how to do a regression test for this.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -28,7 +28,6 @@
                 "mocha/no-exports": "error",
                 "mocha/no-identical-title": "error",
                 "mocha/no-nested-tests": "error",
-                "mocha/no-setup-in-describe": "error",
                 "mocha/no-skipped-tests": "error"
             }
         }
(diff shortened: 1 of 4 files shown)


harrysarson:
@lydell please review 2b50eca :) Otherwise we are good to merge.

lydell:
I think the tests look good! 👍 

Maybe we should disable this lint rule? https://github.com/lo1tuma/eslint-plugin-mocha/blob/master/docs/rules/no-setup-in-describe.md

I don’t understand why it’s complaining

lydell (commented):
Great tests!

harrysarson:
After an embarrassing number of attempts we finally got there

PR #497, opened 2021-02-14
Validate `--fuzz` minimum value earlier

lydell:
Inspired by https://github.com/mpizenberg/elm-test-rs/issues/74, make `--fuzz 0` an error already in CLI parsing, not when running the tests.

Before:

```
❯ elm-test --fuzz 0
Compiling > Starting tests

elm-test 0.19.1-revision6
-------------------------

Running 0 tests. To reproduce these results, run: elm-test --fuzz 0 --seed 347637688619170


TEST RUN INCOMPLETE because Test runner run count must be at least 1, not 0

Duration: 214 ms
Passed:   0
Failed:   0
```

After:

```
❯ elm-test --fuzz 0
error: option '--fuzz <int>' argument '0' is invalid. Expected at least 1.
```

### DIFF
--- a/lib/elm-test.js
+++ b/lib/elm-test.js
@@ -17,12 +17,16 @@ const RunTests = require('./RunTests');
 
 void Report;
 
-const parsePositiveInteger = (string /*: string */) /*: number */ => {
+const parsePositiveInteger = (minimum /*: number */) => (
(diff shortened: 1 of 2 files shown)


harrysarson:
Thanks!

PR #500, opened 2021-03-01
sketch of autopublish action

harrysarson:


harrysarson:
not worth the complexity

Issue #508, opened 2021-05-06
Crash when --output is set for `elm-test make`


juanedi:
Hey! I've been running into some problems when trying to compile Elm tests within Emacs.

The Emacs plugin makes use of the `--output` flag for `elm make` so that users can later use the editor's compilation output ([code here](https://github.com/bsermons/flycheck-elm/blob/master/flycheck-elm.el#L173-L179)). When I try to compile test files (using `elm-test` instead of `elm`) I'm getting `error: unknown option '--output=/dev/null'`.

I first thought about raising this as an issue for flycheck-elm, but after reading https://github.com/rtfeldman/node-test-runner/issues/284 it seemed to me like the intention of `elm-test make` was to be compatible with `elm make` to facilitate editor integration.

I don't know if other editor plugins make use of `--output`, but it didn't seem too weird of a use case so I figured it might make sense to do something here instead of fixing at the plugin level. Would that make sense?

harrysarson:
I am 99% sure that elm-test make never produces any output files. The make subcommand just checks that your tests compile.

Previous we ignored unknown flags but now we raise an error. You can safely delete `--output /dev/null`.

juanedi:
Yeah, I guess I was wondering if `elm-test` should actually ignore that flag when running `make`, with the intention of preserving compatibility with `elm make`.

If we agree that setting `--output` is a reasonable thing for editor plugins to do, then it might make sense to do this instead of expecting every plugin to worry about whether which command is being run and conditionally skip the argument.

TBH that seems like more robust solution to me considering how:

  - AFAIK most editor plugins don't give end users a lot of flexibility for customizing the command and its arguments
  - it might not be trivial for the plugin to detect which program is actually being called
  - `elm-test make` seemed to be meant specifically for editor integration

Thoughts? Thanks for the quick reply!

lydell:
> I am 99% sure that elm-test make never produces any output files.

[...]

PR #510, opened 2021-05-11
Ignore null output option

juanedi:
Fixes https://github.com/rtfeldman/node-test-runner/issues/508

Added a `--output` option for interface compatibility with `elm make`.

Note that:
  - When the option is present, elm-test will fail if anything other than `/dev/null` is passed
  - The option is omitted from `--help`, as suggested by @lydell 
  - I tried logging to stdout and stderr when `--output=/dev/null` is supplied but that didn't play well with my editor plugin. I understand that stderr doesn't work because that's where any plugins will try to read the errors json from, but still would expect stdout to work 🤷 . I didn't look into it in detail, but can do it if you all think it's worth it.

### DIFF
--- a/lib/elm-test.js
+++ b/lib/elm-test.js
@@ -133,14 +133,16 @@ function main() {
       'Use a custom path to an Elm executable (default: elm)',
       undefined
     )
-    .option(
-      "--output <output>",
-      "Allowed for compatibility with `elm make`. The only supported value is /dev/null, which is ignored since elm-test does not produce any output files.",
-      function(value) {
-        if (value !== "/dev/null") {
-         throw new InvalidOptionArgumentError("Only /dev/null is allowed.");
-        }
-      }
+    .addOption(
+      new Option(
+        "--output <output>",
+        "Allowed for compatibility with `elm make`. The only supported value is /dev/null, which is ignored since elm-test does not produce any output files."
+      ).argParser(function(value) {
+          if (value !== "/dev/null") {
+            throw new InvalidOptionArgumentError("Only /dev/null is allowed.");
+          }
+        })
+       .hideHelp()
     )
     .version(packageInfo.version, '--version', 'Print version and exit')
     .helpOption('-h, --help', 'Show help')
--- a/lib/elm-test.js
+++ b/lib/elm-test.js
@@ -133,6 +133,15 @@ function main() {
       'Use a custom path to an Elm executable (default: elm)',
       undefined
     )
+    .option(
+      "--output <output>",
+      "Allowed for compatibility with `elm make`. The only supported value is /dev/null, which is ignored since elm-test does not produce any output files.",
+      function(value) {
+        if (value !== "/dev/null") {
+         throw new InvalidOptionArgumentError("Only /dev/null is allowed.");
+        }
+      }
+    )
     .version(packageInfo.version, '--version', 'Print version and exit')
     .helpOption('-h, --help', 'Show help')
     .addHelpCommand('help [command]', 'Show help');
--- a/lib/elm-test.js
+++ b/lib/elm-test.js
@@ -136,7 +136,7 @@ function main() {
     .addOption(
       new Option(
         "--output <output>",
-        "Allowed for compatibility with `elm make`. The only supported value is /dev/null, which is ignored since elm-test does not produce any output files."
+        "Allowed for compatibility with `elm make`. The only supported value is /dev/null, which elm-test ignores since it does not produce any output files."
       ).argParser(function(value) {
           if (value !== "/dev/null") {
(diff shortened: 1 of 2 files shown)


harrysarson (approved):
LGTM with very minor help message suggestion.

I will let @lydell take a look as they know this code better than I do

lib/elm-test.js
```diff
@@ -133,6 +133,17 @@ function main() {
       'Use a custom path to an Elm executable (default: elm)',
       undefined
     )
+    .addOption(
+      new Option(
+        "--output <output>",
+        "Allowed for compatibility with `elm make`. The only supported value is /dev/null, which elm-test ignores since it does not produce any output files."
```
lydell, line 139: Just curious: Is there any way of making this text show up?

Do we want to document – maybe as a comment – that it’s for Emacs users who use `elm make` for source files, and `elm-test make` for test files, and Emacs-something always adds on `--output /dev/null`?

lib/elm-test.js
```diff
@@ -133,6 +133,17 @@ function main() {
...
+      ).argParser(function(value) {
+          if (value !== "/dev/null") {
+            throw new InvalidOptionArgumentError("Only /dev/null is allowed.");
+          }
+        })
```
lydell, line 144: Another way of writing this is `.choices(['/dev/null'])`. Is it better?

Error message now:

```
error: option '--output <output>' argument 'strawberry' is invalid. Only /dev/null is allowed.
```

Error message with `.choices`:

```
error: option '--output <output>' argument 'strawberry' is invalid. Allowed choices are /dev/null.
```

lydell (commented):
Looks good!

I played around with it a little locally and seems to work like we intended!

@harrysarson Would you like to have a unit test for this (“--output /dev/null should be ignored because …”)

lydell:
I allowed the checks to run now as well. You need to run Prettier on the file you have changed.

lib/elm-test.js
juanedi: > Just curious: Is there any way of making this text show up?

I don't think so 🤔 

> Do we want to document – maybe as a comment – that it’s for Emacs users who use elm make for source files, and elm-test make for test files, and Emacs-something always adds on --output /dev/null?

Will do!

lib/elm-test.js
juanedi: > Another way of writing this is .choices(['/dev/null']). Is it better?

Yup! I totally missed that in the commander docs, but it makes a lot of sense to use the built-in way to validate the attribute.

💭 With that said, I can think of one reason for keeping a custom option, which ties into your previous comment. Since we are omitting the new option in `--help` users might have a hard time figuring out why `--output` behaves the way it does (assuming someone actually notices it!).

With that in mind, we may want to customize the error message a bit to elaborate on why it exists and why it only accepts `/dev/null`.

No strong preference, just thought I'd share the idea before switching to `.choices`.

lydell: I think we can summarize then:

1. Remove the help text for `--output` altogether since it cannot be seen.
2. Use `.argParser` like now with an elaborate error message.
	- Please use `=>` function instead of `function`.
	- This way there’s no need for the “Emacs comment” I asked for – it’ll be in the error message.
3. Duplicate this test and add `--output /dev/null` to it: https://github.com/rtfeldman/node-test-runner/blob/c493674c278e7a2d6f66742b68296af4eb9d8dbd/tests/flags.js#L179-L182
4. Run Prettier

juanedi:
Done!

I took the liberty to add a test for the error case, and left the mention to the emacs package in the comment instead of the expanded docs because I couldn't make it fit in a way that seemed relevant for a non-emacs user running into the error.

No need to say it, but feel free to suggest/make any other modifications to error messages or whatever.

Thanks for the feedback!

lydell:
Looks great! I took the liberty to line-wrap the message to make it easier to read in a wide window. What do you think?

tests/flags.js
```diff
@@ -189,6 +189,28 @@ describe('flags', () => {
...
+      }).timeout(60000);
+
+      it('should fail if setting --output to anything other than /dev/null', () => {
+        const runResult = execElmTest([
+          'make',
+          '--output=output_file',
+          'tests/Passing/One.elm',
+        ]);
+
+        assert.notStrictEqual(runResult.status, 0);
```
harrysarson, line 211:
```suggestion 
        assert.ok(Number.isInteger(runResult.status));
        assert.notStrictEqual(runResult.status, 0);
```

in the past we have had tests passing because `undefined` is not stricty equal to `0`!

harrysarson (approved):
still looks good to me :)

lydell (approved):

PR #558, opened 2021-12-28
Use elm-solve-deps-wasm instead of elm-json solve

mpizenberg:
Hey, I discussed this a little bit over the incremental elm discord, and this is what using `elm-solve-deps-wasm` inside node-test-runner would look like.

[elm-solve-deps-wasm](https://github.com/mpizenberg/elm-solve-deps-wasm) basically is a WebAssembly port of the dependency solver inside elm-test-rs. Advantages over using `elm-json` as a binary are the versatility of the solver that you can customize to your needs (can work with lamdera for example), and lightweight compared to downloading a binary. The wasm crate is 100kb compressed and 280kb total package uncompressed.

On the downsides, it's currently not much tested, has worse error messages than elm-json, and customizability implies a bit more code (cf `dependency-provider-offline.js` provided as starting point).

Let me know what you think of this.


mpizenberg:
PS: it's expected that tests are failing since the starter `dependency-provider-offline.js` file only works if dependencies are already available offline. Tests are passing on my machine when I run `elm-test`.

mpizenberg:
To complement `dependency-provider-offline.js` I've implemented another `dependency-provider.js` that contains both the offline and online logic, and basically reproduces the full logic of the solver in elm-test-rs in JS. I think I got it right but JS is so painfully unforgiving that there are probably some bugs left.

I've left untouched the `dependency-provider-offline.js` because it's much easier to read first, to get a grasp of the API, and then read the more complete `dependency-provider.js`, which has additional online behavior and also caches.

Since I haven't figured out (yet) how to interact with async functions and wasm, I've made the wasm interfaces sync, and thus needed to have sync versions of file reading and http requests. This means I've added `sync-request` to the dependencies. If you know how to prevent that, let me know.

mpizenberg:
I think I've figured out a way to use async functions with wasm and rust, but unfortunately, due to the contaminating nature of async, it requires changes all the way through elm-solve-deps to the pubgrub rust package. And there are other WIP in pubgrub that are more prioritized than async, so I wouldn't expect to be able to use `elm-solve-deps-wasm` in an async way for a few months.

lib/DependencyProvider.js
```diff
@@ -0,0 +1,317 @@
+// @flow
+
+const fs = require('fs');
+const os = require('os');
+const path = require('path');
+const request = require('sync-request').default;
```
lydell, line 6: TODO(lydell): Audit if we want this dependency. It seems pretty heavy (2MB): https://packagephobia.com/result?p=sync-request. @mpizenberg what does it do?

lib/DependencyProvider.js
```diff
@@ -0,0 +1,317 @@
...
+    fs.writeFileSync(cachePath, elmJson);
+    return elmJson;
+  }
+}
+
+function fetchElmJsonOffline(
+  pkg /*: string */,
+  version /*: string */
+) /*: string */ {
+  console.log('Fetching: ' + pkg + ' @ ' + version);
```
lydell, line 36: Are these `console.log`s temporary?

If we want to log something, I think it should be part of the `Compiling > Solving dependencies > Starting tests` “progress bar”. I noticed “Solving dependencies” has been removed. But this approach still makes network requests, right? If so, isn’t “Solving dependencies” still needed since it can take time?

lib/DependencyProvider.js
```diff
@@ -0,0 +1,317 @@
...
+// This is to be cleared before each call to solve_deps().
+const listVersionsMemoCache /*: Map<string, Array<string>> */ = new Map();
+
+function fetchElmJsonOnline(
+  pkg /*: string */,
+  version /*: string */
+) /*: string */ {
+  try {
+    return fetchElmJsonOffline(pkg, version);
+  } catch (_) {
```
lydell, line 21: Is it safe to not care about what error we caught? 🤔 I haven’t followed the whole control flow through yet. (Same thing in a couple of other places.)

lydell (commented):
I pushed some modifications directly to your branch, I hope that’s ok!

lib/DependencyProvider.js
```diff
@@ -0,0 +1,317 @@
...
+const os = require('os');
+const path = require('path');
+const request = require('sync-request').default;
+
+// Cache of existing versions according to the package website.
+let onlineVersionsCache /*: Map<string, Array<string>> */ = new Map();
+
+// Memoization cache to avoid doing the same work twice in listAvailableVersions.
+// This is to be cleared before each call to solve_deps().
+const listVersionsMemoCache /*: Map<string, Array<string>> */ = new Map();
```
lydell, line 13: We already have a cache in Solve.js. Can you explain how this cache comes into play?

mpizenberg:
I haven't looked at the cache within solve.js
There are multiple caches inside this implementation. OnlineCacheVersion is
a hash map that corresponds to the Json answered by the package server at
the address /packages. For efficiency this is also stored on disk in all
home inside a pubgrub/ directory. Like so, when in online mode, we can
start by reading this file on dis, count package versions, and and only ask
the server the new packages at packages/since/count.

There is a temporary cache just for memoization of the answer to the
function listing existing  versions for a given package. This is just to
avoid having to read the disk again to list installed versions and merge
with existing version according to the package website. (This merge of
local + server answer is to have compatibility with lamdera).

Finally, I also cache the JSON of each package when needing to read the
dependencies of a given package. Both in a memory cache and on disk inside
the pubgrub subdirectory inside the elm home.

lydell:
The way the cache in Solve.js works is this:

1. Try to read `elm-stuff/generated-code/elm-community/elm-test/0.19.1-revision6/dependencies.SHA256.json` from disk.
2. If it exists, use that (it contains the `"dependencies"` field value for the generated Elm application used to run the tests).
3. Otherwise compute that.

Reading that json file is very fast.

The cache key is a SHA256 hash of the user’s elm.json, so it is only invalidated if the user changes their elm.json which usually isn’t that often.

I know that `elm-json` stores things in `~/.elm`, but I’m not sure how that affects things the times `elm-json` is actually invoked.

Given the above, do you think the extra caching you built is needed? If so, can you give an example scenario?

mpizenberg:
Ah yes it's a completely different level of caching. And I think all levels
are useful. Elm-json employs a similar caching strategy. It's just that the
nature of a dependency solver may need to ask some of the things I'm
caching multiple times.

lydell:
So, what’s an example scenario where the extra caching leads to a faster experience?

lib/DependencyProvider.js
mpizenberg: So sync-request is here because it's not possible yet to have async callback functions in the pubgrub solver (Follow this https://github.com/pubgrub-rs/pubgrub/issues/110). So for now, both `fetchElmJson` and `listAvailableVersions` must be sync functions. I didn't know how to make sync http requests in Node so I found this package that enable sync http requests. If there is another way to make sync http requests don't hesitate to change this.

lib/DependencyProvider.js
mpizenberg: Yes the logs are temporary, definitely better to have proper logging. I think I removed the logging because I wanted to remove the logging related to downloading elm-json, I don't remember for sure but I had the impression that the logging logic for solving dependencies and for downloading elm-json was tied so I removed all of it.

In practice, with the few caches, the solver should almost never make http requests. Typically, there will be some requests the first time a package is seen by the solver on the user machine. Then, the cache in `~/.elm/pubgrub/` will be enough to solve everything offline, even in other project folders since the cache does not live in `elm-stuff/`. But it's probably better to re-add a `Solving dependencies` step?

lib/DependencyProvider.js
mpizenberg: There aren't many ways the functions in the `try` can fail. It may be better to differentiate between failures that are normal, and failure that are unexpected, but for this proof of concept, there is no need to do so. We are basically redoing the work in online if the offline mode failed, so there isn't much risk for now. Can be improved later in this PoC.

mpizenberg:
> So, what’s an example scenario where the extra caching leads to a faster experience?

There are few scenarios I have in mind such as:

## Create another project with the same dependencies than the current project

Let's say you use some project as a starter for another project. Change a few things like the package name, or the source location. But the dependencies stay unchanged. All files needed to solve the dependencies will already be available offline, so no need to make a request to the package server.

## Add a dependency

With the caches, most of what is needed is already on disk. Most of the time, only few new requests will be needed instead of 1 + 1 per dependency. Sometimes even 1 request will be enough.

If the offline solver fails, there is one mandatory request done to `https://.../all-packages/since/${count}` where `count` is the number of package versions already known. Which are the ones we got last time we did a `.../all-packages/since/${lastCount}` request. So this requests will be quite fast as it just lists the last few published packages since last time we updated this. It will look something like follows.

```js
// curl -L https://package.elm-lang.org/all-packages/since/12970 | jq
[
  "bChiquet/elm-accessors@2.1.1",
  "newmana/chroma-elm@18.0.6",
  "newmana/chroma-elm@18.0.5"
]
```

Then only when needing to fetch the dependencies of a given package for the first time, inside `fetchElmJson` we need to make a call to the package server like `https://package.elm-lang.org/packages/elm/random/1.0.0/elm.json`. But sometimes, that package elm json was already downloaded for solving dependencies in another project so no request is needed at all.

lib/DependencyProvider.js
harrysarson: I would like to architect this such that:

1. `DependancyProvider` is a class
2. We create a new `DependancyProvider` for each call to `solve_deps()` rather than clearing a global variable.
3. We pass (bound?) methods from the `DependancyProvider` class to 

```
    const dependancyProvider = new DependencyProvider();
    
    // Solve again, in online mode.
    solution = wasm.solve_deps(
      JSON.stringify(elmJson),
      useTest,
      extra,
      () => dependancyProvider.fetchElmJsonOnline(),
      () => dependencyProvider.listAvailableVersionsOnline(),
    );
```

lydell: (Note that last time I tried to use a class, I couldn’t make Flow and Prettier work with it.)

harrysarson:
I think I like the direction of this, using a library (even if it is wasm) is much neater than shelling out to a different program. It has the green light from me in principle.

lib/DependencyProvider.js
```diff
@@ -0,0 +1,317 @@
...
+  let versionsCount = 0;
+  for (const versions of onlineVersionsCache.values()) {
+    versionsCount += versions.length;
+  }
+
+  // Complete cache with a remote call to the package server.
+  const remoteUrl = remotePackagesUrl + '/since/' + (versionsCount - 1); // -1 to check if no package was deleted.
+  const newVersions = JSON.parse(request('GET', remoteUrl).getBody('utf8'));
+  if (newVersions.length === 0) {
+    // Reload from scratch since it means at least one package was deleted from the registry.
```
harrysarson, line 96: what happens if a package is deleted and then a different package is immediately uploaded?

mpizenberg:
> what happens if a package is deleted and then a different package is
immediately uploaded?

In normal conditions, we request at since/${count-1} such that we also have
the latest package we know of in the answer from the server. As such, we
can verify that no package was removed.

Let's say the last package we know is A.

If A is deleted, and another package B is added, we will find B where we
expected A in the response and deduce that something not normal happened.
So we reset known versions.

If another package than A is deleted, and a new package C is added, it's
the same. Since C is more recent than A, it's C that will be the package at
the frontier and we will get C where we expected A in the response. In this
situation, we also reset packages.

In general, if any number of packages we knew are deleted, either there
will be an empty response, or, if new packages were added, there will be
the wrong package at position ${count} compared to what we were expecting,
and in these situations, we reset known packages.


If packages that we don't know yet are deleted before we update known
packages, it has no impact on our logic. It will be as if they never
existed. Other packages depending on those will just fail to resolve due to
no version existing for that dependency.

lib/DependencyProvider.js
lydell: I think I’d experiment replacing it with this approach (similar to how I do requests in `elm-tooling`):

1. Try to spawn `curl` (exists on macOS, Windows, some Linux) using [child_process.spawnSync](https://nodejs.org/api/child_process.html#child_processspawnsynccommand-args-options)
2. If `curl` doesn’t exist (ENOENT), try to spawn `wget` (exists on some Linux) using using [child_process.spawnSync](https://nodejs.org/api/child_process.html#child_processspawnsynccommand-args-options)
3. If none of them exist, spawn `node fallback-http.js` using [child_process.spawnSync](https://nodejs.org/api/child_process.html#child_processspawnsynccommand-args-options)

A benefit of using `curl` and `wget` is that people can then configure those to use proxies etc. And it should be much more lightweight.

mpizenberg: Your approach is better I think indeed. I don't plan on coding much until next week. If there is an easy way to reuse the elm-tooling approach don't hesitate to change it if you have the time. I can give it a try next week otherwise.

lib/DependencyProvider.js
mpizenberg: One remark, the cache for memoization should be reset between calls, because otherwise you can't take into account updates of existing versions. But the cache for online versions doesn't need to be reset between solver calls, just updated (if we want) with `updateOnlineVersionsCache`. This can be a bit faster in the context of watch mode with changes in the `elm.json` invalidating the checksum cache. Not a lot faster, but a little bit, since it avoids the JSON file read and de-serialization into a hashmap. And just need to do the hashmap -> json -> write updated file serialization after the update is done.

lib/DependencyProvider.js
mpizenberg: Note for my future self, look at how we can make a sync request using a worker thread and Atomics: https://whistlr.info/2021/block-nodejs-main-thread/

lydell: Not sure if that’s easier/better than what I suggested, but feel free to try!

mpizenberg: I've figured out a way to do this that seems to work base on this minimal example: https://github.com/mpizenberg/sync-get

However, when I do this inside node-test-runner, it's hanging. I haven't figured out yet why the request doesn't terminate and notify the main thread. I'll dig a little bit.

mpizenberg: I've pushed to the branch in case someone else wants to help diagnose the hanging issue.

mpizenberg: Alright I found the issue which turned out to be a very stupid issue of module paths. I had to replace `new Worker("./SyncGetWorker.js"` by `workerPath = path.resolve(__dirname, "SyncGetWorker.js");` and `new Worker(workerPath`

So good news, the dependency was removed, and by the way, this `SyncGetWorker` pattern might interest you @lydell (cf https://github.com/mpizenberg/sync-get).

What do you think?

PS: also fixed a bug due to forgetting to `JSON.parse` the get request for `elm.json` files to the package server.

lib/DependencyProvider.js
mpizenberg: What do you think of something like this:
```js
  try {
    const offlineDepsProvider = new DependencyProvider.Offline();
    solution = offlineDepsProvider.solve(elmJson, useTest, extra);
  } catch (_) {
    // console.error('Offline solver failed, switching to new online');
    const syncGetWorker = SyncGet.startWorker();
    const onlineDepsProvider = new DependencyProvider.Online(syncGetWorker);
    solution = onlineDepsProvider.solve(elmJson, useTest, extra);
    syncGetWorker.shutDown();
  }
```
So inside the `Online` and `Offline` classes of the `DependencyProvider` module, we expose a `solve()` function that internally does the clearing of the memoization cache and uses the correct versions of `fetchElmJson` and `listAvailableVersions`. The online versions cache could be updated at the beginning of the `solve()` call, or only at the constructor initialization and then exposing an update function for the cache in the online class.

Here I'm also passing the object enabling synchronous GET requests based on a worker discussed in another thread here. This is because it's better to create the worker only once instead of one time per solve call, to avoid doing it in watch mode. But for simplicity, we could also do this directly inside the `Online` class. I don't think the perf change would be big anyway cause starting a worker shouldn't be too expensive I think. Especially considering that the tests runner already creates new child processes for the runners.

mpizenberg:
I've updated the code on two points discussed above.

First, I've removed the dependency to `sync-request`. Instead I've used a technique consisting in using directly the `https` module in a worker thread. Inside the `DependencyProvider` module, I initialize the worker by calling `SyncGet.startWorker()` that starts a worker and initialize an object with a `get` method for GET requests. Then everytime I need to make a synchronous request inside the dependency provider, I call `syncGetWorker.get(url)` which performs the requests and returns the body of the response. Internally, this send a message to the worker containing the url, which performs the series and async calls necessary, and then wake up the main thread with `Atomics.notify()`. At this point we get the body response and can continue synchronously as normal in our main thread. I think this approach is lightweight enough since it just start one NodeJS worker thread, which is rather cheap once Node is already started. Then all subsequent `.get(url)` calls just use the same worker. So this might even be more efficient that the approach using curl / wget to perform requests synchronously.

Second, I've moved the cache logic directly within the DependencyProvider module, now exposing only two constructor functions `newOffline()` and `newOnline()`, returning an object with a single `solve()` method performing automatically the cache operations required.

I've added and modified some flow declaration to have `npm run flow` passing. However `eslint` is failing, saying that `SharedArrayBuffer` and `Atomics` are not defined. If you know how to fix that, don't hesitate.

There still console logs here and there coming from the wasm module, so this is not perfect yet, but we are much closer from something really usable. So now I'll wait for the next round of reviews for feedback.

lydell:
Looks like the tests are failing because the `worker_threads` module is behind the `--experimental-worker` flag in Node.js 10. But it should be fine to drop Node.js 10 now (it’s been EOL for almost a year). We dropped support for Node.js 8 last year with my big PRs. Maybe ideally dropping Node.js 10 should be its own PR (we have some code with comments about cleanups to do when dropping Node.js – also in this PR I think btw).

mpizenberg:
I think the 10.x test run is failing because node 10.x can only run worker threads with the `--experimental-worker` command line argument. I don't know if that is very important. Probably depends if support for Node 10.x is kept in elm-test-rs next release.

Could we add `fail-fast: false` to the test strategy to check other versions?

harrysarson:
> But it should be fine to drop Node.js 10 now (it’s been EOL for almost a year). 

absolutely agree

lydell:
I made a PR to drop Node.js 10: https://github.com/rtfeldman/node-test-runner/pull/559

mpizenberg:
So, if we want to continue making progress on this PR after the one dropping node 10 is done, here are the things left to do.

1. Catch specific errors that are expected to fail an offline solver, and re-throw errors that are unexpected.
2. Add compatibility with elm 0.19.0.
3. Fix the transfer of a solver error from wasm to JS.

(1) As discussed in a previous review, the current code has some catch-all `catch (_)` branches in places where we are expecting failures with fallback strategies. This is for example the case if the offline solver fails, leading to an online solver attempt. So it is expected that the offline solver fails if a package is not installed, preventing us to pick any version, or to read some `elm.json` files. However, if the offline solver were to fail for an unexpected reason, we might want to stop and report that error instead of switching to the online solver. So here we need to make expected failures more explicit in the code.

(2) Currently, to list installed versions and to retrieve dependencies of installed versions, the code will look into `~/.elm/0.19.1/packages/`. We should instead have the version depend on the current elm compiler. The structure is the same for elm 0.19.0 but I don't know about the structure of this directory for earlier versions like 0.18 and 0.17. Is elm-test currently compatible with elm versions older than 0.19.0?

(3) I've recently removed direct calls to `console.log` from within the wasm code to prevent it from messing with the `elm-test` output. Normally the error message should be available in the `catch (error)` of the JS code calling the failing wasm code. However, it seems this is not the case, so I might have to investigate why that is. And if I cannot figure out why, I'll probably return something like `{ ok: ... }` or `{ err: ... }` directly in the wasm output, and pattern match on this in the JS code.

## Things not worth the effort in my opinion

The current code has removed logging indicators that dependency solving is running. But the solver has a caching strategy, consisting in storing on disk inside `~/.elm/` the `elm.json` files needed to read dependencies of required packages. Since that cache inside `~/.elm/` is permanent and can be used across projects, there is a very low probability that the solver will need to reach for network, and thus the solver will very likely spend less than a few milliseconds to run. This is not true in the following situations:

- the `~/.elm/` cache is removed
- a package is required for the first time since last time `~/.elm/` was cleared (and thus has not even been installed by a regular `elm make` call before running tests)

In my opinion, those situations only happen on CI, or for `lamdera reset` users. Re-adding an indicator depends on how much weight we give to these use cases.

## Things that can be improved later

Current error messages are not tailored for elm and elm-test, but rather the generic errors provided by pubgrub when solving fails. I think they are sufficiently explicit for users to understand what is going on and can be improved on later. Here is an example showcasing the different error messages when running. First the one of the wasm solver:

```txt
Because there is no version of rtfeldman/elm-hex in 2.0.0 <= v < 3.0.0
and avh4/elm-color 1.0.0 depends on rtfeldman/elm-hex 2.0.0 <= v < 3.0.0,
avh4/elm-color 1.0.0 is forbidden.
```

Then the one given by elm-json:

```txt
Because rtfeldman/elm-hex 2.0.0 <= v < 3.0.0 is unavailable
and this project depends on rtfeldman/elm-hex 2.0.0 <= v < 3.0.0,
no valid set of package versions could be found.
```

This is another one, for an app instead of a package this time. First with the wasm solver.

```txt
Because there is no version of rtfeldman/elm-hex in 2.0.0
and root/ 0.0.0 depends on rtfeldman/elm-hex 2.0.0,
root/ 0.0.0 is forbidden.
```

And now with elm-json.

```txt
Because rtfeldman/elm-hex 2.0.0 is unavailable
and this project depends on rtfeldman/elm-hex 2.0.0,
no valid set of package versions could be found.
```

In my opinion, improving error messages should be done directly in the rust lib `elm-solve-deps` which is used by the wasm solver, so that everyone can take advantage of the improvements (like elm-test-rs also depending on this, and potential other tooling reusing the rust or wasm version of the lib). Alternatively, we could return directly the structured tree required to generate the error message, and would rewrite a reporting logic fully inside JS. I don't think this is a wise use of our coding time.

I don't think of anything else for now.

avh4:
The versioning scheme for elm-test npm package thus far has been to have a separate release for each elm version.  Looking at https://www.npmjs.com/package/elm-test -> Versions, shows the latest releases and number of downloads in the past week:
- `0.19.1-revision7` 9,539 downloads
- `0.19.0-rev6` 40 downloads
- `0.18.13` 4,827 downloads
- `0.17.3` 0 downloads

For an improvement like this, imo it would be sufficient to only apply it to the 0.19.1 version, and leave the older series as they are.

lib/DependencyProvider.js
```diff
@@ -0,0 +1,350 @@
...
+
+// We can replace this with using `Object.fromEntires` once Node.js 10 is
+// EOL 2021-04-30 and support for Node.js 10 is dropped.
+function fromEntries(entries) {
+  const res = {};
+  for (const [key, value] of entries) {
+    res[key] = value;
+  }
+  return res;
+}
```
lydell, line 226: This should be removed now and we can use `Object.fromEntries()` instead since has been merged

mpizenberg: I'll remove it thanks!

mpizenberg:
1. [x] Catch specific errors that are expected to fail an offline solver, and re-throw errors that are unexpected.
2. ~Add compatibility with elm 0.19.0.~
3. [x] Fix the transfer of a solver error from wasm to JS.

(1) I've improved error handling with more fine grained `try` blocks. Most often, they now contain a single instruction that we expect can fail. There is a big one left though, the main one actually. This is the one in `Solve.js`:

```js
try {
  const offlineDepsProvider = DependencyProvider.newOffline();
  solution = offlineDepsProvider.solve(JSON.stringify(elmJson), useTest, extra);
} catch (_) {
  // console.error('Offline solver failed, switching to online');
  const onlineDepsProvider = DependencyProvider.newOnline();
  solution = onlineDepsProvider.solve(JSON.stringify(elmJson), useTest, extra);
}
```

I think this is fine though because all steps of the offline dependency provider are also done inside the online dependency provider. So if something unexpected and fatal is to happen in the offline solver, it should also happen in the online solver and thus be reported.

(2) I didn't add 0.19.0 compat since we said it's fine.

(3) This was just that I need to convert the string error from the wasm failure into an `Error` by re-throwing it so that it has a `.message` field which is used in elm-test elsewhere.

-------------

There is still one bug I just found. At two locations I'm sorting and/or reversing a list of versions where the order is not correct. Indeed versions are always in their String form in this JS code, and the string order of versions is wrong if we want to get newest versions first. This is because "10." is sorted before "2.". Is there another place in elm-test code base where versions are handled more correctly as semver instead of as strings? If I can copy paste or reuse something already there would be nice.

This bug will not prevent solver to work, but will not return the expected newest versions in some circumstances.

lydell:
> This is because "10." is sorted before "2."

Not sure if it helps, but sorting using an [Intl.Collator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator) you can set the `numeric` option to sort numbers-in-strings numerically (examples in the article).

mpizenberg:
Actually it's possible to pass a compare function to the sort function so I did just that.

Right now I don't think of anything else left to change so I'm waiting for another round of reviews. To me the last thing needed is to ask people to try it and see if things break unexpectedly.

lydell:
Was that easier than using collator? At a quick glance, it looks like you might be able to replace 50 lines of code with 1.

mpizenberg:
I had a quick look and didn't find what I was looking for. Then I looked at
the compatibility table down the page and saw Node 13 so I just went for
the compare function.

lydell:
This program runs even on Node.js 6 (that’s the oldest version I had installed), and seems to give the same results as your function:

```js
const collator = new Intl.Collator('en', { numeric: true });

function compareStringVersions(
  v1 /*: string */,
  v2 /*: string */
) /*: number */ {
  const [major1, minor1, patch1] = v1.split('.');
  const [major2, minor2, patch2] = v2.split('.');
  let left = parseInt(major1);
  let right = parseInt(major2);
  if (left < right) {
    return -1;
  } else if (left > right) {
    return 1;
  } else {
    left = parseInt(minor1);
    right = parseInt(minor2);
    if (left < right) {
      return -1;
    } else if (left > right) {
      return 1;
    } else {
      left = parseInt(patch1);
      right = parseInt(patch2);
      if (left < right) {
        return -1;
      } else if (left > right) {
        return 1;
      } else {
        return 0;
      }
    }
  }
}

function randomNumber() {
  return Math.floor(100 * Math.random());
}

function randomVersion() {
  return `${randomNumber()}.${randomNumber()}.${randomNumber()}`;
}

while (true) {
  const version1 = randomVersion();
  const version2 = randomVersion();
  console.log(`${version1} <=> ${version2}`);
  const x = compareStringVersions(version1, version2);
  const y = collator.compare(version1, version2);
  if (x !== y) {
    console.log(`${x} !== ${y}`);
    break;
  }
}
```

Would you mind the shorter version?

mpizenberg:
Thanks, yes if it works the same the collator version is better

mpizenberg:
@lydell I have slightly modified the `parseVersions(key, json)` function to avoid an unnecessary duplication of the `json` object when it is already a valid array of versions. But now Flow is complaining about the types not being correct. I've tried a few things but didn't figure out a type annotation fix. If I say that `json` is already an `Array<string>` it complains about it not being `mixed` and if I say it's `mixed`, it complains about the output not being an `Array<string>` but a `mixed` even though we did a dynamic check. Is that why you had created an additional array?

Flow is also complaining about the `collator.compare` being "unbound from the context where it was defined". I'm not sure what this means. Do you know how to fix that?

lydell:
> Is that why you had created an additional array?

Yes, that feels like the only way of doing these things in my experience (it’s easier to create a value with a known type than to refine a `mixed` type). I wouldn’t worry about it – duplicating the array shouldn’t be time consuming?

mpizenberg:
> duplicating the array shouldn’t be time consuming?

The `parseVersion(key, json)` function is called on all packages inside the response from `https://.../all-packages/` (or equivalent file on disk). So it accumulates to allocating roughly 13000 entries as of today (there are 13000 versions total). I'd prefer to avoid doing that allocation twice, since it's already done by the `JSON.parse(getResponseOrFileOnDisk)` we give as input.

I've added the following Flow annotation to make it trust us instead. I hope that's ok.
```js
  // $FlowFixMe[incompatible-return]: We dynamically checked that `json` is an `Array<string>`.
  return json;
```

lydell:
Then I’m very curious – how long does building that new array take with a typical JSON file?

mpizenberg:
> Then I’m very curious – how long does building that new array take with a typical JSON file?

I think you're right. I've tried benchmarking the difference and it's negligeable for the current size of this file. If the file grows 100x to 1 300 000 entries, then the additional allocations lead to a bump from 185ms to 235ms to build the Map, so even if Elm package base grows 100x it will mean a 50ms increase. I'll let you decide what to do about that.

lydell:
> I'll let you decide what to do about that.

It probably doesn’t matter. It feels like the format of that file is very unlikely to change, and therefore the code is unlikely to change too so the typechecking won’t save us from anything. On the other hand, it sounds like a premature optimization. I don’t know 🤷‍♂️

mpizenberg:
Yeah, this file probably won't change until the scale of the Elm language forces Evan to take another approach. Let's leave it as is, and wait for @harrysarson point of view.

Any other remarks or does this PR look good enough to you @lydell to try it IRL?

lib/DependencyProvider.js
```diff
@@ -0,0 +1,367 @@
...
+
+// Cache of existing versions according to the package website.
+let onlineVersionsCache /*: Map<string, Array<string>> */ = new Map();
+
+// Memoization cache to avoid doing the same work twice in listAvailableVersions.
+// This is to be cleared before each call to solve_deps().
+const listVersionsMemoCache /*: Map<string, Array<string>> */ = new Map();
+
+// Constructor of an offline solver.
+function newOffline() /*: {
```
lydell, line 23: Why does this (and `newOnline`) return an object with the “real” function? Couldn’t we expose `solveOffline` which would be the internal `solve` function directly?

lib/DependencyProvider.js
```diff
@@ -0,0 +1,367 @@
...
+    fs.writeFileSync(cachePath, elmJson);
+    return elmJson;
+  }
+}
+
+function fetchElmJsonOffline(
+  pkg /*: string */,
+  version /*: string */
+) /*: string */ {
+  // console.log('Fetching: ' + pkg + ' @ ' + version);
```
lydell, line 90: Can we remove all logging leftovers now? (Search for `// console`)

mpizenberg: Ah I forgot about those. I wanted to ask if there is a logging mechanism in elm-test which can be enabled with configurable verbosity but I forgot. If not, we can just remove all. If yes, we could add some of those commented logs to that.

lib/DependencyProvider.js
mpizenberg: You're totally right. I initially did this because I started with a class design but could not only expose what I wanted, all was public. So I replaced the class by a function returning an object. And then ended up removing all things I had put in the object except the actual useful function and kept the global variables of the module since it was more convenient.

In the end only the function stays so we can definitely peel that layer!

lydell:
I ran this PR on the test suite we have at work. Didn’t spot any issues! It was really cool that the offline solver succeeded on the first try, even with no `~/.elm/pubgrub/` folder existing! Hadn’t realized it could read from `~/.elm/0.19.1/packages/` before (even though my eyes must have went over that code at some point). Cool stuff!

Next thing we can try is to see if anyone on Slack or Discord would like to try this PR out. They can install like so:

```
npm install --save-dev mpizenberg/node-test-runner#elm-solve-deps-wasm
```

mpizenberg:
> Hadn’t realized it could read from ~/.elm/0.19.1/packages/ before

That's what enables compatibility with lamdera or any other compiler using the same structure than the elm compiler.

lib/DependencyProvider.js
lydell: We don’t have a logging mechanism, unfortunately

mpizenberg:
I've removed the commented `// console.log()` lines and now expose directly `solveOffline()` and `solveOnline()` functions in the dependency provider as suggested by Simon.

mpizenberg:
Nitpick question, do you prefer
```js
  let solution;
  const elmJsonStr = JSON.stringify(elmJson);
  try {
    solution = DependencyProvider.solveOffline(elmJsonStr, useTest, extra);
  } catch (_) {
    solution = DependencyProvider.solveOnline(elmJsonStr, useTest, extra);
  }
  return solution;
```
or with `return` statements inside?
```js
  const elmJsonStr = JSON.stringify(elmJson);
  try {
    return DependencyProvider.solveOffline(elmJsonStr, useTest, extra);
  } catch (_) {
    return DependencyProvider.solveOnline(elmJsonStr, useTest, extra);
  }
```
I'm wondering if I should shave those 2 lines to have it slightly more consise and readable? even if I try to avoid multiple return statements when not needed.

lydell:
I prefer the second version with two `return` statements and no `let`.

mpizenberg:
Note to self. Check what exactly is the minimum required version with current code. According to @avh4 it seems some of the worker stuff for the sync http request needs node a version newer than 12.20. I need to have a look at that and see if some can be changed or not.

harrysarson:
I made a PR against this branch: https://github.com/mpizenberg/node-test-runner/pull/1

Hopefully it tidies up the global state a bit.

mpizenberg:
Thanks Harry! I'll try to review next week. Can you explain the gist of the changes?

harrysarson:
updated the PR description. Let me know if there is more I can say!

lydell:
What is left to do in this PR?

- Check minimum required Node.js 12 version: https://github.com/rtfeldman/node-test-runner/pull/558#issuecomment-1018117862? Alternatively, wait one month and Node.js 12 will be end of life so we can drop it instead? https://github.com/nodejs/release
- Somehow get people to test on big projects? I guess we could at least try to get Jeroen to do it, since he’ll be interested in this elm-solve-deps-wasm strategy for elm-review.
- @harrysarson Do you have more review you would like to do?

I’m excited about this PR and don’t feel good about leaving it hanging, but I have other things I’d like to focus on for the moment.

avh4:
FYI, I have a large app I can test it on.

Also, re node 12, imo I think it'd be fine to just drop it now, so long as there's a clear error message explaining that you need node 14+ if you try to run it on an earlier version (or does npm even have a way to mark the package as requiring a certain node version?)  Anyone committed to using node 12 can just lock to the current elm-test version.

lydell:
> re node 12, imo I think it'd be fine to just drop it now, so long as there's a clear error message explaining that you need node 14+ if you try to run it on an earlier version (or does npm even have a way to mark the package as requiring a certain node version?)

Yes, package.json has the `engines` field: https://github.com/rtfeldman/node-test-runner/pull/465#discussion_r515549375

mpizenberg:
I don't think there is much left to discuss since the discussion happened on the other PR by @harrysarson . Now I guess we only need a bit of testing from people with big code base.
After that, for Jeroen (and others) it might be nice to extract the JS overlay on top of the wasm package that we ended up with here so that there is a JS package with just a couple function ready to be used as-is. Otherwise everyone will end with a slightly different JS overlay on top of the wasm package. We might want it for flexibility but it's more update and fix maintainability burden if issues are raised.

mpizenberg:
One argument to keep compatibility with node 12 is that debian bullseye, which is [the current stable debian uses Node 12.22](https://packages.debian.org/search?searchon=names&keywords=nodejs).

I have had issues raised about node 10 compatibility for elm-test-rs for the previous stable debian, so it might become an issue for elm-test too if we remove node 12 too early. Locking an elm-test version is an option, but it's possible they need to update for a security patch. I don't know, might be simpler to release a security patch for an older release.

I think all should work with a minimum version of 12.20 (for the sync http requests via worker) so if we can keep it the minimum version let's keep it that way.

harrysarson:
agree about node 12. We should update the minimum version in package.json (currently it is 12.13 I think) and possibly also ensure that our CI checks node 12.20

mpizenberg:
I added `12.20.0` to the list of node versions tested in the `test.yml` action. Not sure if that works since I can't trigger CI myself. Maybe there is also a way to extract that from the node engine property but I haven't looked at how if possible.

.github/workflows/test.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest]
-        node-version: [12.x, 14.x, 16.x, 17.x]
+        node-version: [12.20.0, 12.x, 14.x, 16.x, 17.x]
```
lydell, line 16: Is it worth testing both 12.20.0 and latest 12.x, or would it be enough with just 12.20.0?

mpizenberg: Felt like both the explicit minimum version, and the likely version are worth checking but :shrug:

avh4:
Trying this out on my work project(s) with `npm install -g mpizenberg/node-test-runner#elm-solve-deps-wasm`:

Everything seems to work when using npm 6.14.16!

But when using npm 8.5.5, `npm install -g mpizenberg/node-test-runner#elm-solve-deps-wasm` fails with:
```ascii
npm ERR! npm ERR! code 127
npm ERR! npm ERR! path /root/.npm/_cacache/tmp/git-cloneXJGWfj
npm ERR! npm ERR! command failed
npm ERR! npm ERR! command sh -c elm-tooling install
npm ERR! npm ERR! sh: 1: elm-tooling: not found
npm ERR!
npm ERR! npm ERR! A complete log of this run can be found in: ...
```

Seemingly when trying to run the `prepare` hook?

avh4:
ah yeah, that analysis sounds correct then.  I'm not seeing any npm hooks that would be appropriate for this (post-install script that is only for when dev dependencies are installed).  Maybe the solution is to change the prepare script to be `if which elm-tooling; then elm-tooling install; fi` ?

mpizenberg:
> Maybe the solution is to change the prepare script to be if which elm-tooling; then elm-tooling install; fi ?

That would not be windows-compatible would it?

lydell:
@avh4 You try adding `--ignore-scripts` to your `npm install` command.

avh4:
Oh, npm scripts still don't get `sh` on Windows?

I'll see if `--ignore-scripts` works; that would be fine for me, since it's gonna be in a Dockerfile.  (Also, if it weren't for building a docker image, I wouldn't be using global install anyway.)

lydell:
But it’s still a good point of `prepare`. If npm 8 runs `prepare` on global installs, that’s bad. People are going to do that. So I should look into running elm-tooling for development in this repo some other way (and update the elm-tooling docs).

mpizenberg:
So it seems we have the green light to merge, and the elm-tooling prepare stuff can be fixed later in another PR since it's already impacting the current version of node-test-runner.

Anything else you want to check @harrysarson ?

harrysarson:
@mpizenberg do you want to draft a commit message for the squashed commit?

mpizenberg:
Use a custom dependency solver library instead of the elm-json binary.

This PR introduces a custom dependency solver, extracted from the solver in elm-test-rs, improving the compatibility with other elm-based platforms such as Lamdera.

The core algorithmic part of the solver, called "pubgrub", is rooted in the same logic than used in elm-json but implemented differently in the rust package `pubgrub-rs`.
That core algorithmic logic has been ported into a WebAssembly package called `elm-solve-deps-wasm` dedicated to the constraints of the elm package ecosystem and providing the following `solve_deps` wasm function where the two functions passed as arguments `js_fetch_elm_json` and `js_list_available_versions` are two functions injected via the JS call.

```rust
pub fn solve_deps(
    project_elm_json_str: &str,
    use_test: bool,
    // additional_constraints_str: &HashMap<String, Constraint>,
    additional_constraints_str: JsValue,
    // js_fetch_elm_json(pkg: &str, version: &str) -> String;
    js_fetch_elm_json: js_sys::Function,
    // js_list_available_versions(pkg: &str) -> Vec<String>;
    js_list_available_versions: js_sys::Function,
) -> Result<JsValue, JsValue> {
```

In this PR, the JS overlay implementing those two `js_fetch_elm_json` and `js_list_available_versions` is taken care of in the form of a `DependencyProvider` class providing two functions `solveOffline` and `solveOnline`. In the case of the offline solver, listing of existing versions and discovery of dependencies is exclusively performed from the packages installed inside the elm home. In the case of the online solver, some information may be retrieved from the package server.

```js
// same interface for solveOnline
solveOffline( 
  elmJson /*: string */,
  useTest /*: boolean */,
  extra /*: { [string]: string } */
) /*: string */ { ... }
```

Remark: since requests for the online solver need to perform http requests in a synchronous manner, an approach based on a web worker thread is used, requiring a minimum node version of 12.20.0.

mpizenberg:
Above is an attempt at a summary of this PR for a commit message. Feel free to change/remove/add anything.

harrysarson:
Thanks!

PR #592, opened 2022-05-10
Add a command to use master version of elm-explorations/test

Janiczek:
This is based on a [discussion](https://discord.com/channels/534524278847045633/762656477608542228/971472576167673907) in the Incremental Elm #elm-test Discord channel.

The intent is to allow people to help us beta-test the `elm-explorations/test` 2.0.0 release and use the `master` version of `elm-explorations/test` on their codebases (masqueraded as the 1.2.2 version).

cc @harrysarson 

https://user-images.githubusercontent.com/149425/167733507-6aa73ed7-0a30-4102-9d15-26a3fc891ef2.mp4

### DIFF
--- a/lib/Install.js
+++ b/lib/Install.js
@@ -11,6 +11,36 @@ const Project = require('./Project');
 
 void Project;
 
+function rmDirSync(dir) {
+  // We can replace this with just `fs.rmSync(installationScratchDir, { recursive: true, force: true })`
+  // when Node.js 12 is EOL 2022-04-30 and support for Node.js 12 is dropped.
+  // `fs.rmSync` was added in Node.js 14.14.0, which is also when the
+  // `recursive` option of `fs.rmdirSync` was deprecated. The `if` avoids
+  // printing a deprecation message.
+  // $FlowFixMe[prop-missing]: Flow does not know of `fs.rmSync` yet.
+  if (fs.rmSync !== undefined) {
+    fs.rmSync(dir, { recursive: true, force: true });
+  } else if (fs.existsSync(dir)) {
+    // $FlowFixMe[extra-arg]: Flow does not know of the options argument yet.
+    fs.rmdirSync(dir, { recursive: true });
+  }
+}
+
+function rmFileSync(file) {
+  // We can replace this with just `fs.rmSync(installationScratchDir, { recursive: true, force: true })`
+  // when Node.js 12 is EOL 2022-04-30 and support for Node.js 12 is dropped.
+  // `fs.rmSync` was added in Node.js 14.14.0, which is also when the
+  // `recursive` option of `fs.rmdirSync` was deprecated. The `if` avoids
+  // printing a deprecation message.
+  // $FlowFixMe[prop-missing]: Flow does not know of `fs.rmSync` yet.
+  if (fs.rmSync !== undefined) {
+    fs.rmSync(file, { recursive: false, force: true });
+  } else if (fs.existsSync(file)) {
+    // $FlowFixMe[extra-arg]: Flow does not know of the options argument yet.
+    fs.rmdirSync(file, { recursive: false });
+  }
+}
+
 function install(
   project /*: typeof Project.Project */,
   pathToElmBinary /*: string */,
@@ -22,18 +52,7 @@ function install(
     // Recreate the directory to remove any artifacts from the last time
     // someone ran `elm-test install`. We do not delete this directory after
     // the installation finishes in case the user needs to debug the test run.
-    // We can replace this with just `fs.rmSync(installationScratchDir, { recursive: true, force: true })`
-    // when Node.js 12 is EOL 2022-04-30 and support for Node.js 12 is dropped.
-    // `fs.rmSync` was added in Node.js 14.14.0, which is also when the
-    // `recursive` option of `fs.rmdirSync` was deprecated. The `if` avoids
-    // printing a deprecation message.
-    // $FlowFixMe[prop-missing]: Flow does not know of `fs.rmSync` yet.
-    if (fs.rmSync !== undefined) {
-      fs.rmSync(installationScratchDir, { recursive: true, force: true });
-    } else if (fs.existsSync(installationScratchDir)) {
-      // $FlowFixMe[extra-arg]: Flow does not know of the options argument yet.
-      fs.rmdirSync(installationScratchDir, { recursive: true });
-    }
+    rmDirSync(installationScratchDir);
     fs.mkdirSync(installationScratchDir, { recursive: true });
   } catch (error) {
     throw new Error(
@@ -125,7 +144,7 @@ async function downloadFileNative(url, filePath) {
     const file = fs.createWriteStream(filePath);
     let fileInfo = null;
 
-    const request = https.get(url, response => {
+    const request = https.get(url, (response) => {
       if (response.statusCode !== 200) {
         fs.unlink(filePath, () => {
           reject(new Error(`Failed to get '${url}' (${response.statusCode})`));
@@ -144,11 +163,11 @@ async function downloadFileNative(url, filePath) {
     // The destination stream is ended by the time it's called
     file.on('finish', () => resolve(fileInfo));
 
-    request.on('error', err => {
+    request.on('error', (err) => {
       fs.unlink(filePath, () => reject(err));
     });
 
-    file.on('error', err => {
+    file.on('error', (err) => {
       fs.unlink(filePath, () => reject(err));
     });
 
@@ -156,35 +175,37 @@ async function downloadFileNative(url, filePath) {
   });
 }
 
-async function installUnstableTestMaster(project) {
+async function installUnstableTestMaster(
+  project /*: typeof Project.Project */
+) {
   const pkg = 'elm-explorations/test';
   const version = '1.2.2';
 
-  console.log(chalk.yellow(`Using the master version of ${pkg} in place of ${version}.`));
-  console.log(chalk.yellow(`Note you will need to rm -rf ~/.elm and ./elm-stuff (or use the \`elm-test uninstall-unstable-test-master\` command) afterwards to get back to the ${version} version.`));
+  console.log(
+    chalk.yellow(`Using the master version of ${pkg} in place of ${version}.`)
+  );
+  console.log(
+    chalk.yellow(
+      `Note you will need to rm -rf ~/.elm and ./elm-stuff (or use the \`elm-test uninstall-unstable-test-master\` command) afterwards to get back to the ${version} version.`
+    )
+  );
 
-  const pkgWithDash = pkg.replace('/','-');
+  const pkgWithDash = pkg.replace('/', '-');
 
   const tempPath = project.generatedCodeDir;
   const zipballUrl = `https://codeload.github.com/${pkg}/zip/refs/heads/master`;
   const zipballFilename = `${pkgWithDash}.zip`;
   const zipballPath = path.join(tempPath, zipballFilename);
 
-  const packagePath = path.join(
-    elmHome,
-    '0.19.1',
-    'packages',
-    pkg,
-    version
-  );
+  const packagePath = path.join(elmHome, '0.19.1', 'packages', pkg, version);
 
   console.log(chalk.dim.yellow(`Removing ${tempPath}`));
-  fs.rmSync(tempPath, { recursive: true, force: true });
+  rmDirSync(tempPath);
 
   console.log(chalk.dim.yellow(`Removing ${packagePath}`));
-  fs.rmSync(packagePath, { recursive: true, force: true });
+  rmDirSync(packagePath);
 
-  fs.mkdirSync(tempPath,    { recursive: true });
+  fs.mkdirSync(tempPath, { recursive: true });
   fs.mkdirSync(packagePath, { recursive: true });
 
   console.log(chalk.dim.yellow(`Downloading ${zipballUrl}`));
@@ -196,7 +217,8 @@ async function installUnstableTestMaster(project) {
     [
       '-o', // overwrite
       zipballFilename, // file to unzip
-      '-d', tempPath // directory where to extract files
+      '-d',
+      tempPath, // directory where to extract files
     ],
     {
       cwd: tempPath,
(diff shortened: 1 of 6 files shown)


Janiczek:
From Discord:

> I think we should use a different elm home but is not a strong feeling

> I think there's a lot of other small things to think about, like more alerting color for the warning, and cross-compatible ways to do the shell commands, description of the flag, etc...

gampleman:
I wonder if this could even take an argument, so that it could be used to test locally other branches?

Janiczek:
@gampleman My initial assumption when hearing @harrysarson's idea was that this would later be reverted, and not a long-term flag. If it was to be longterm, I agree customizing it could be a nice addition.

lib/Install.js
```diff
@@ -117,6 +118,44 @@ function install(
...
+          reject(error);
+        }
+        resolve([stdout,stderr]);
+      });
+    });
+
+  await ourExec(`rm -rf ${tempDir}`);
+  await ourExec(`mkdir -p ${tempDir}`);
+  await ourExec(`rm -rf ${packagePath}`);
+  await ourExec(`rm -rf elm-stuff/generated-code`);
```
lydell, line 147: Should this only delete `elm-stuff/generated-code/elm-community/elm-test`? To avoid deletin elm-review’s generated code for example.

Also, all paths should be written based on the “root dir” like in https://github.com/rtfeldman/node-test-runner/blob/fd52f043b9ae6ce20596af123af3be1a7967a383/lib/Project.js#L46-L53

lib/Install.js
```diff
@@ -117,6 +118,44 @@ function install(
...
+  await ourExec(`mkdir -p ${packagePath}`);
+  await ourExec(`curl -L ${zipballUrl} -o ${zipballPath}`);
+  await ourExec(`unzip -o ${zipballPath} -d ${tempDir}`);
+  await ourExec(`mv ${tempDir}/test-master/src ${packagePath}`);
+  await ourExec(`mv ${tempDir}/test-master/elm.json ${packagePath}`);
+  await ourExec(`mv ${tempDir}/test-master/LICENSE ${packagePath}`);
+  await ourExec(`mv ${tempDir}/test-master/README.md ${packagePath}`);
+  await ourExec(`rm -rf '${zipballPath}'`);
```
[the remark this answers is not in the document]
Janiczek: Given this comment
```js
    // We can replace this with just `fs.rmSync(installationScratchDir, { recursive: true, force: true })`
    // when Node.js 12 is EOL 2022-04-30 and support for Node.js 12 is dropped.
```
I feel compelled to just use the simpler one 😁

lib/Install.js
```diff
@@ -19,18 +24,8 @@ function install(
...
-    // `fs.rmSync` was added in Node.js 14.14.0, which is also when the
-    // `recursive` option of `fs.rmdirSync` was deprecated. The `if` avoids
-    // printing a deprecation message.
     // $FlowFixMe[prop-missing]: Flow does not know of `fs.rmSync` yet.
-    if (fs.rmSync !== undefined) {
-      fs.rmSync(installationScratchDir, { recursive: true, force: true });
-    } else if (fs.existsSync(installationScratchDir)) {
-      // $FlowFixMe[extra-arg]: Flow does not know of the options argument yet.
-      fs.rmdirSync(installationScratchDir, { recursive: true });
-    }
```
lydell, line 33: As I said in an earlier comment, it was decided to support Node.js 12 for the time being, so this needs to be reverted so CI can pass.

It’s fine though if the new feature is Node.js 14+ (as long as CI passes).

lib/elm-test.js
```diff
@@ -241,12 +245,12 @@ function main() {
...
     // command and only run tests in `src/`, ignoring all files in `tests/`.
     .command('__elmTestCommand__ [globs...]', { hidden: true, isDefault: true })
-    .action((testFileGlobs) => {
+    .action(async (testFileGlobs) => {
       const options = program.opts();
       const pathToElmBinary = getPathToElmBinary(options.compiler);
       const projectRootDir = getProjectRootDir('tests');
       const processes = Math.max(1, os.cpus().length);
-      RunTests.runTests(
+      await RunTests.runTests(
```
lydell, line 253: Why was this change made? `.action()` does not take a function that returns a promise (`async` functions always return promises). Instead we run the promise and handle the results with `.then` and `.catch`. In general, one needs `.catch()` at the top-most level of your program, where your promise chain ends. Otherwise if the promise fails you’ll end up with an uncaught promise rejection.

lib/elm-test.js
```diff
@@ -165,12 +169,12 @@ function main() {
...
     .command('init')
     .description('Install elm-explorations/test and create tests/Example.elm')
-    .action(() => {
+    .action(async () => {
       const options = program.opts();
       const pathToElmBinary = getPathToElmBinary(options.compiler);
       const project = getProject('init');
       try {
-        Install.install(project, pathToElmBinary, 'elm-explorations/test');
+        await Install.install(project, pathToElmBinary, 'elm-explorations/test');
```
lydell, line 177: Does `Install.install` return a promise now? See the other comment about `async`/`await`. I’d extract most stuff here to a helper function I think, where I can use `async`/`await`, and then call it here with `.then` and `.catch` like the other command.s

lydell (commented):
TODO:

- Make CI pass
- Test manually on Windows (I think `spawn.sync` throws an error rather than returning a non-zero exit code if `unzip` does not exist)

harrysarson (approved):
I am happy to delegate detail review to others, the shape of this PR looks great!

Given this is a tool to test the new elm-explorations which should be uniform across platforms (its elm) I would think we _need_ windows support (although it would obviously be cool to have).

harrysarson:
> I wonder if this could even take an argument, so that it could be used to test locally other branches?
@[Janiczek](https://github.com/Janiczek) Janiczek commented [4 days ago](https://github.com/rtfeldman/node-test-runner/pull/592#issuecomment-1123841837)
@gampleman My initial assumption when hearing @harrysarson's idea was that this would later be reverted, and not a long-term flag. If it was to be longterm, I agree customizing it could be a nice addition.

I was thinking of  this being something we revert, but if we end up liking I could see us building it out

lib/elm-test.js
Janiczek: `.action()` does seem to take a function that returns a promise.

https://github.com/tj/commander.js/blob/6c26370fcba737d9786a34d0e8693f689a067569/lib/command.js#L1237
:arrow_down:
https://github.com/tj/commander.js/blob/6c26370fcba737d9786a34d0e8693f689a067569/lib/command.js#L1146-L1154

lib/elm-test.js
Janiczek: Good catch, thanks! I previously implemented this hacking of elm-explorations/test inside Install.install before realizing that's the wrong place, and didn't fully revert the changes to Install.install.

Reverted in ae25add

lib/Install.js
Janiczek: Reverted in 0d5d3c2

lib/elm-test.js
lydell: Hmm, yes, that seems to be the case! I found this:

https://github.com/tj/commander.js#action-handler

> You may supply an `async` action handler, in which case you call `.parseAsync` rather than `.parse`.

But we use `.parse`. I wonder if I had a reason for that when adding `commander` 1.5 years ago, or if I simply did not know about `.parseAsync`.

Janiczek: Should I change to `parseAsync` then?

lydell: Maybe! I would have to play around with it locally to learn how it works.

lib/Install.js
```diff
@@ -117,6 +122,122 @@ function install(
...
+  const version = '1.2.2';
+  const pkgWithDash = pkg.replace('/','-');
+
+  const tempPath = path.join(
+    projectRootDir,
+    'elm-stuff',
+    'generated-code',
+    'elm-community',
+    'elm-test'
+  );
```
lydell, line 178: If you take a `Project` as a parameter instead of just the project root dir, you can use `project.generatedCodeDir` instead of repeating this.

lib/Install.js
```diff
@@ -117,6 +122,122 @@ function install(
...
+    }
+  }
+  fs.renameSync(
+    path.join(
+      tempPath,
+      'test-master'
+    ),
+    packagePath
+  );
+  fs.rmSync(zipballPath, { recursive: true, force: true });
```
lydell, line 237: This does not need to be recursive, does it?

lydell:
@Janiczek Let me know if you need help getting `flow`, `eslint` or `prettier` to pass.

lib/elm-test.js
lydell: I played around a bit locally.

- You can remove `async` and `await` here – it’s not needed.
- I think we could simplify some code by using `async` action handlers and `parseAsync`, but that’s out of scope for this PR.

Janiczek: > You can remove async and await here – it’s not needed.

Any reasoning for this? I'm using async/await in the `Install.installUnstableTestMaster` function. Why is it not needed in its usages?

Janiczek: Ah, I've missed the `.then(exit(0),exit(1))` afterwards. Makes sense!

Janiczek:
This now works via a pair of **commands:**
* `elm-test install-unstable-test-master`
* `elm-test uninstall-unstable-test-master`

lib/elm-test.js
```diff
@@ -142,6 +142,10 @@ function main() {
       '--color',
       'Force colored console output (setting FORCE_COLOR to anything but 0 also works)'
     )
+    .option(
+      '--unstable-use-test-master',
+      'Use the bleeding-edge version of the elm-explorations/test library'
```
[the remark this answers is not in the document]
Janiczek: I've went with a pair of commands in the end.

lydell:
@Janiczek I fixed the CI failure, but since you used your `master` branch I don’t seem to be able to push it…

lib/Install.js
```diff
@@ -117,6 +124,129 @@ function install(
...
+
+async function installUnstableTestMaster(
+  project /*: typeof Project.Project */
+) {
+  console.log(
+    chalk.yellow(`Using the master version of ${PKG} in place of ${VERSION}.`)
+  );
+  console.log(
+    chalk.yellow(
+      `Note: You will need to rm -rf ~/.elm and ./elm-stuff (or use the \`elm-test uninstall-unstable-test-master\` command) afterwards to get back to the ${VERSION} version.`
```
lydell, line 174: One last question: I’m wondering if people will wonder “Hmm, which approach should I take? Delete two folders, or run the black box command?” Do we want to mention _only_ `elm-test uninstall-unstable-test-master`?

Janiczek: Since we're saying what we're doing during these commands (with the `chalk.yellow.dim` stuff), I think we can remove the mention of `rm -rf` here, yes. I'll do that

lydell:
This is looking really good now!

I finally tested on Windows, and unfortunately I ran into an error:

```
PS C:\Users\Simon\stuff\node-test-runner\example-application> node ..\bin\elm-test install-unstable-test-master
Using the master version of elm-explorations/test in place of 1.2.2.
Note: You will need to use the `elm-test uninstall-unstable-test-master` command afterwards to get back to the 1.2.2 version.
Removing C:\Users\Simon\stuff\node-test-runner\example-application\elm-stuff\generated-code\elm-community\elm-test\0.19.1-revision7
Removing C:\Users\Simon\AppData\Roaming\elm\0.19.1\packages\elm-explorations\test\1.2.2
Downloading https://codeload.github.com/elm-explorations/test/zip/refs/heads/master
Unzipping C:\Users\Simon\stuff\node-test-runner\example-application\elm-stuff\generated-code\elm-community\elm-test\0.19.1-revision7\elm-explorations-test.zip
Moving to ELM_HOME: C:\Users\Simon\AppData\Roaming\elm\0.19.1\packages\elm-explorations\test\1.2.2
EPERM: operation not permitted, rename 'C:\Users\Simon\stuff\node-test-runner\example-application\elm-stuff\generated-code\elm-community\elm-test\0.19.1-revision7\test-master' -> 'C:\Users\Simon\AppData\Roaming\elm\0.19.1\packages\elm-explorations\test\1.2.2'
PS C:\Users\Simon\stuff\node-test-runner\example-application>
```

Not sure why that happens yet, but I don’t have more time right now.

lydell:
I figured it out! Made another PR here: https://github.com/Janiczek/node-test-runner/pull/2/files

lib/Install.js
```diff
@@ -117,6 +124,122 @@ function install(
...
+      fileInfo = {
+        mime: response.headers['content-type'],
+        size: parseInt(response.headers['content-length'], 10),
+      };
+
+      response.pipe(file);
+    });
+
+    // The destination stream is ended by the time it's called
+    file.on('finish', () => resolve(fileInfo));
```
lydell, line 149: We seem to resolve the promise with some data we never use here.

lib/Install.js
```diff
@@ -117,6 +124,122 @@ function install(
   return 'SuccessfullyInstalled';
 }
 
+async function downloadFileNative(url, filePath) {
```
lydell, line 127: Maybe we should add type annotations to all new functions in this file – for parameters and return types.

lydell:
One _final_ final question 😅 What way is the easiest to verify that:

- I actually get the master version.
- I was able to back to the published version.

lib/Install.js
lydell: Went ahead and did the last final touches: https://github.com/Janiczek/node-test-runner/pull/3

lydell:
Replying to myself, I figured out an easy way to test: I added `Debug.log "a" Fuzz.andThen` and checked if the tests resulted in a compilation error or a print. It works!

However, I first thought it didn’t work, because I tried with `example-application/` in this repo, and it used elm-test 1.2.<strong>0</strong> so nothing happened. Is it worth doing anything about that? Checking the version in elm.json (sounds annoying to code). Wait – I just noticed our prints _do_ mention that we’re only replacing version 1.2.2 so maybe that’s good enough? Especially since 1.2.2 was released [June 29, 2019](https://github.com/elm-explorations/test/releases/tag/1.2.2) so all _real_ tests out there probably use the latest version anyway?

<details>

<summary>Here’s where 1.2.2 is mentioned multiple times:</summary>

```
❯ ../bin/elm-test install-unstable-test-master
Using the master version of elm-explorations/test in place of 1.2.2.
Note: You will need to use the `elm-test uninstall-unstable-test-master` command afterwards to get back to the 1.2.2 version.
Removing /Users/lydell/forks/node-test-runner/example-application/elm-stuff/generated-code/elm-community/elm-test/0.19.1-revision7
Removing /Users/lydell/.elm/0.19.1/packages/elm-explorations/test/1.2.2
Downloading https://codeload.github.com/elm-explorations/test/zip/refs/heads/master
Unzipping /Users/lydell/forks/node-test-runner/example-application/elm-stuff/generated-code/elm-community/elm-test/0.19.1-revision7/elm-explorations-test.zip
Moving to ELM_HOME: /Users/lydell/.elm/0.19.1/packages/elm-explorations/test/1.2.2
Removing /Users/lydell/forks/node-test-runner/example-application/elm-stuff/generated-code/elm-community/elm-test/0.19.1-revision7/elm-explorations-test.zip
```

</details>

Janiczek:
Good to see you figured how to test this out:
> What way is the easiest to verify that:
> I actually get the master version.
> I was able to back to the published version.

But anyways just to post my way of testing this, I've simply ran this local version of elm-test in one of my repos that had `Shrink.noShrink`.
* `1.2.2` will show no error
* `master` will say the `Shrink` module doesn't exist

So it's usually fine to do something like
```bash
cd ~/Localhost/elm/elm-secret-sharing
../../cloned/node-test-runner/bin/elm-test # 1.2.2, works
../../cloned/node-test-runner/bin/elm-test install-unstable-test-master
../../cloned/node-test-runner/bin/elm-test # master, doesn't work
../../cloned/node-test-runner/bin/elm-test uninstall-unstable-test-master
../../cloned/node-test-runner/bin/elm-test # 1.2.2, works
```

lydell:
I noticed that the intellij-elm plugin does not like when I run `elm-test install-unstable-test-master`. It asks me to “re-attach elm.json”, and when I do it gives an error. Is that something you’ve noticed as well?

Janiczek:
> I noticed that the intellij-elm plugin does not like

I don't use that one (I'm pretty light on Elm editor plugins) so I didn't notice 😬 
I don't suppose intellij-elm checks hashes / redownloads archives, so perhaps this is just a matter of stale cache? Would the `File | Invalidate Caches` button help?

lydell:
Found the issue and made a PR: https://github.com/Janiczek/node-test-runner/pull/4

Janiczek:
Those auto-close comments are treacherous 😁

lydell (approved):

Issue #581, opened 2022-04-12
Implicitly exposed tests not found if test module contains a character literal with a unicode sequence


nonpop:
**In short**

If your test file contains a character literal like `'\u{000D}'`, then no tests are found in it unless you explicitly expose them by name. The string `"\u{000D}"` is not a problem an neither is the character `'\r'`. The sequence 000D is not special; other sequences such as `'\u{0041}'` hide the tests, too.

**Details**

elm version 0.19.1, elm-test version 0.19.1-revision7

- `elm.json`:
```json
{
    "type": "application",
    "source-directories": [
        "src"
    ],
    "elm-version": "0.19.1",
    "dependencies": {
        "direct": {
            "elm/core": "1.0.3",
            "elm/json": "1.1.3"
        },
        "indirect": {}
    },
    "test-dependencies": {
        "direct": {
            "elm-explorations/test": "1.2.2"
        },
        "indirect": {
            "elm/html": "1.0.0",
            "elm/random": "1.0.0",
            "elm/time": "1.0.0",
            "elm/virtual-dom": "1.0.2"
        }
    }
}
```
- Empty `src/` directory
- `tests/Foo.elm`:
```elm
module Foo exposing (..)

import Expect
import Test exposing (..)

foo = test "foo" <| \() -> Expect.pass
```
- Run `elm-test`. Tests pass
- Now add `c = '\u{000D}'` at the end of the Foo module (could be any sequence)
- Run `elm-test`. Fails since no tests are found.
- Expose `foo` explicitly: `module Foo exposing (foo)`. Tests pass again
    - Tests also pass if you expose both `foo` and `c` explicitly
    - They also pass if you replace \u{000D} with the equivalent \r
    - They also pass if instead of a character you have a string: `c = "\u{000D}"`

lydell:
Awesome bug report! Char literals with unicode escapes are indeed not accounted for in the parser: https://github.com/rtfeldman/node-test-runner/blob/fd52f043b9ae6ce20596af123af3be1a7967a383/lib/Parser.js#L546-L562

PR #594, opened 2022-05-14
Fix parsing of char literals with unicode escapes

lydell:
Fixes https://github.com/rtfeldman/node-test-runner/issues/581

Note: I had a hard time deciding how “well” we should parse this. elm-test doesn’t care at all what’s inside strings and char literals – it just needs to “get past them” without being confused about later syntax. I think I ended up with a good trade-off in the end. See the code comments in the diff for more details.

### DIFF
--- a/lib/Parser.js
+++ b/lib/Parser.js
@@ -266,7 +266,10 @@ const backslashableChars = new Set([
   '"',
   "'",
   '\\',
-  // `u` must be followed by `{1234}` but we don’t bother.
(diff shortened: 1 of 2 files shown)


fredcy:
I tested this version of the test runner against the problematic code that I reported on Slack (test containing `|> Fuzz.map (String.cons '\u{1F648}')`) and it does fix the problem in that case. So it looks good to me. Nice work!

PR #599, opened 2022-05-29
Remove reversing step for offline versions

jfmengels:
We were sorting then reversing the list in 2 locations. It's more efficient to sort in the opposite order once.

(cc @mpizenberg who did the original implementation)

### DIFF
--- a/lib/DependencyProvider.js
+++ b/lib/DependencyProvider.js
@@ -115,7 +115,7 @@ class OnlineAvailableVersionLister {
     for (const version of offlineVersions) {
       allVersionsSet.add(version);
     }
-    const allVersions = [...allVersionsSet].sort(semverCompare).reverse();
+    const allVersions = [...allVersionsSet].sort(semverCompare);
     this.memoCache.set(pkg, allVersions);
     return allVersions;
   }
@@ -149,11 +149,7 @@ function readVersionsInElmHomeAndSort(pkg /*: string */) /*: Array<string> */ {
     offlineVersions = [];
   }
 
-  offlineVersions.sort(semverCompare);
-
-  // Reverse order of subdirectories to have newest versions first.
-  offlineVersions.reverse();
-  return offlineVersions;
+  return offlineVersions.sort(semverCompare);
 }
 
 class DependencyProvider {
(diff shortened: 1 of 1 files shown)


lib/DependencyProvider.js
```diff
@@ -258,7 +254,7 @@ function onlineVersionsFromScratch(
 // Helper functions ##################################################
 
 function semverCompare(a /*: string */, b /*: string */) /*: number */ {
-  return collator.compare(a, b);
+  return -collator.compare(a, b);
```
jfmengels, line 257: This is also possible. I'm fine with either implementation, but I feel like using `-` makes it slightly more obvious that we're sorting in the opposite order.

```suggestion
  return -collator.compare(b, a);
```

lydell: If this function (`semverCompare`) was written in Elm, we would probably have returned an `Order`. Following [Basics.compare](https://elm.dmy.fr/packages/elm/core/latest/Basics#compare), I would expect:

- `LT` if `a` is less than `b` (semver wise)
- `EQ` if they are the same
- `GT` if `a` is greater than `b`

In JavaScript, my understanding is:

- LT = -1
- EQ = 0
- GT = 1

So now this function returns the opposite of what one would expect (because that’s what we need where it’s used now)? Should we rename in to `flippedSemverCompare`? If so, is `collator.compare(b, a)` better? 🤔

jfmengels: I'm fine with that. I changed it to what you suggested and added a small explanation.

mpizenberg:
I don't think the reverse has noticeable impact on the performances. The function was kept in its default order because it's possible in the future to use the increasing order if we want to have an option that picks older versions first. I have such option in elm-test-rs and it's convenient to check that the lower bounds of a package dependencies are correct.

But I don't mind if you want to change the sorting order, I don't have any strong opinion about this.

mpizenberg:
Yes, I didn't mean to say it's changing anything. I just meant to say that I initially preferred having one `semverCompare` function and use it for older versions, or combine with a reversing for newer versions. But in practice, we don't have an option yet to solve with older versions in priority. So it makes sense to use a reverse order sorting instead.

lydell (approved):

PR #612, opened 2022-06-20
Fix install-unstable-test-master for packages

lydell:
@jfmengels reported this error when trying `elm-test install-unstable-test-master` on a _package:_

```
npx elm-test install-unstable-test-master
Could not find "elm-explorations/test": "1.2.2" in your elm.json file here:

/home/jeroen/dev/elm-review-unused/elm.json

This command only works if you have elm-explorations/test as a (direct) test-dependency,
and only if you use version 1.2.2.

You seem to be using version 1.2.2 <= v < 2.0.0.
```

For some reason, Elm uses version _ranges_ event for `test-dependencies` for packages. I assumed test-dependencies were always single versions, and never tested packages.

This PR adds supports for packages, in a cheat way. Rather than implementing range calculations, we simply require the exact string `"1.2.2 <= v < 2.0.0"`: (edit: we also allow whitespace differences)

- 1.2.2 was released [June 29, 2019](https://github.com/elm-explorations/test/releases/tag/1.2.2) (three years ago), so people are likely to have that version range.
- If they don’t, it should be ok to have users change to it.

### DIFF
--- a/lib/Install.js
+++ b/lib/Install.js
@@ -169,28 +169,41 @@ function getDirectTestDependencies(
 
 const PKG = 'elm-explorations/test';
 const VERSION = '1.2.2';
+const VERSION_RANGE = '1.2.2 <= v < 2.0.0';
+
+function getExpectedVersion(
+  project /*: typeof Project.Project */
+) /*: string */ {
+  switch (project.elmJson.type) {
+    case 'application':
+      return VERSION;
+    case 'package':
+      return VERSION_RANGE;
+  }
+}
 
 async function installUnstableTestMaster(
   project /*: typeof Project.Project */
 ) /*: Promise<void> */ {
   const directTestDependencies = getDirectTestDependencies(project);
   const actualVersion = directTestDependencies[PKG];
-  if (actualVersion !== VERSION) {
+  const expectedVersion = getExpectedVersion(project);
+  if (actualVersion !== expectedVersion) {
     throw new Error(
       `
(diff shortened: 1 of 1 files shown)


harrysarson:
Looks good, one question: what if users have the string `"1.2.2 <= v <  2.0.0"` (note the extra space)? Could the error message be confusing for them?

lydell:
Elm is super strict about the spaces:

```
❯ elm make
Dependencies ready!
-- PROBLEM WITH CONSTRAINT -------------------------------------------- elm.json

I got stuck while reading your elm.json file. I do not understand this version
constraint:

13|         "elm/project-metadata-utils": "1.0.0  <= v < 2.0.0"
                                                 ^
I need something like "1.0.0 <= v < 2.0.0" that explicitly lists the lower and
upper bounds.

Note: The spaces in there are required! Taking them out will confuse me. Adding
extra spaces confuses me too. I recommend starting with a valid example and just
changing the version numbers.
```

`elm-json` doesn’t like the extra space either:

```
-- INVALID ELM.JSON ------------------------------------------------------------

Invalid range: 1.2.0  <= v < 2.0.0
```

However, adding an extra space in "test-dependencies" seems to be fine – I think it’s because our dependency solver is not strict about spaces. (The generated elm.json gets something like `"elm-explorations/test": "1.2.2"` in it.)

So at first I was going to answer that it won’t happen, but now I’m not quite sure. I guess we could remove all whitespace from both actual and expected before comparing if we want.

PR #614, opened 2022-08-26
Fuzzer coverage statistics / elm-explorations/test 2.0.0

Janiczek:
Edit by @lydell:

I’ve updated the PR to basically do everything that’s needed for the elm-explorations/test 2.0.0 release.

TODO when 2.0.0 has been released:

- [x] Resolve the TODO in README.md
- [x] Make elm-review pass
- [x] Make tests pass (hopefully they already do!)

---

Adds support for https://github.com/elm-explorations/test/pull/187. It's not merged yet but it's assumed it will be a part of the 2.0.0 release.

The shape of `report` that `Supervisor.js` consumes changes a bit for the Console reporter: we no longer distinguish between passes and failures by `null` vs `string`, but we have a similar object structure to the JSON runner: `status` field with `'pass' | 'todo' | 'fail'`.

Here it can be seen running (with `npm run elm-test`).

Screenshot 2022-08-27 at 0 05 32

Screenshot 2022-08-27 at 0 06 25

The expected behaviour (from my side, in the console runner) is:
* NoCoverage -> nothing gets shown
* CoverageToReport -> coverage table gets shown in both pass and failure scenario
* CoverageCheckSucceeded -> not shown
* CoverageCheckFailed -> shown as part of the failure message (thus outside of the Console reporter jurisdiction)


Janiczek:
(The tests do not pass because this needs to run against the `statistics` branch of `elm-explorations/test`.

lydell:
Does that mean that there’s incompatibility between the package and runners?

|  | node-test-runner 0.19.1-revision9 | node-test-runner this PR |
|-|-|-|
| elm-explorations/test 1.2.2 | ✅ | ❓ |
| elm-explorations/test 2.0.0 | ❓ | ✅ |

If there is incompatibility:

- Is _possible_ to have compatibility?
- How hard/worth would that be?
- What could we do to help users get the right combo?

Janiczek:
The runner Elm code now needs the function `Test.Runner.getCoverageReport`. Adding this function could be a minor/patch change, if I'm not mistaken. The rest of the changes in the `statistics` PR, like changing the `FuzzOptions` record, would likely be a majir change.

So we _might_ have a way to make a new version of `node-test-runner` compatible with a new `elm-explorations/test` 1.x.x version. Is that worth it though?

I believe an old version of `node-test-runner` might be API-compatible with `elm-explorations/test` 2.0.0 (hopefully - haven't checked whether you use `TestOptions` etc.), it just will not report the coverage stuff that it should. I don't think it's a good state to be in long-term.

It seems to me like we should synchronize releases and encourage people to upgrade node-test-runner to the new version if they want to use elm-explorations/test 2.0.0 to the full potential.

harrysarson:
re compatibility I think the new version of the node test runner does need to support both elm-explorations/test versions. However, we can get the user to help by providing `--use-legacy-test-v1`. Just because major upgrades of elm packages can be tricky (say they use some testing package that hasn't upgraded yet).

lydell:
Can the runner detect which version of elm-explorations/test you are using? Then we could print an error like “You use elm-explorations/test vX. This version of the elm-test CLI only supports elm-explorations/test vY”

harrysarson:
> Can the runner detect which version of elm-explorations/test you are using? Then we could print an error like “You use elm-explorations/test vX. This version of the elm-test CLI only supports elm-explorations/test vY”

this is easy right? You just `JSON.parse` the `elm.json`

lydell:
Oh, yeah, didn’t think of that :) We even have code that (kind of) does that for the `elm-test unstable-install-test-master` command:

https://github.com/rtfeldman/node-test-runner/blob/026581eb144a32defd6dae3c3de5c46e5e987527/lib/Install.js#L170-L210

However – it is a bit hacky for packages, which have a _range_ like `1.2.2 <= v < 2.0.0`. But maybe it’s enough if the range string starts with `2.` for the purpose I suggested above (detecting compatibility). So that could work out!

Janiczek:
I wonder, if we wanted compatibility of new runner with old library, would we have to keep essentially two runner sources and dispatch to one or the other based on the version?

Janiczek:
In a873267 I've added tests to `ci.js` for when the coverage table should/shouldn't show up, for all 3 reporters and all interesting coverage cases.

Screenshot 2022-09-24 at 15 40 05

There were also a few changes needed to make existing tests pass with the future-2.0.0: 3505780 (mainly `Expect.true`, `Expect.false` and `Fuzz.tuple3` needed to be migrated)

----

The main issue for me now is that this depends on `elm-explorations/test` future-2.0.0 being installed as 1.2.2 by the person running the tests. (ie. as if `elm-test install-unstable-test-master` was being run with the `statistics` branch instead of `master`)

But I don't know how to go forwards until we formulate a plan for the releases. Longterm I believe it makes sense to change these `elm.json` dependencies to say `elm-explorations/test: 2.0.0`. Should I do that now? Should we run things locally with faked `2.0.0` release to see everything lines up?

Other than that the PR is ready now I believe. No further changes intended from my side.

elm/elm.json
```diff
@@ -6,20 +6,25 @@
...
         "direct": {
-            "elm-explorations/test": "1.2.0"
+            "elm-explorations/test": "2.0.0"
         },
         "indirect": {
-            "elm/html": "1.0.0",
-            "elm/virtual-dom": "1.0.2"
+            "elm/bytes": "1.0.8",
+            "elm/random": "1.0.0",
+            "turboMaCk/queue": "1.1.0"
```
lydell, line 27: Are all these changes intentional?

- elm-review flags for example `elm/browser` as unused.
- Does it work to move some depenencies from direct to indirect like that?

elm/tests/Test/Reporter/Console/FormatTest.elm
```diff
@@ -58,4 +60,58 @@ suite =
...
+                Test.reportCoverage
+                    [ ( "fizz", \n -> (n |> modBy 3) == 0 )
+                    , ( "buzz", \n -> (n |> modBy 5) == 0 )
+                    , ( "even", \n -> (n |> modBy 2) == 0 )
+                    , ( "odd", \n -> (n |> modBy 2) == 1 )
+                    ]
+            }
+            (Fuzz.intRange 1 20)
+            "Fizz buzz even odd - fail"
+            (\n -> Expect.fail "boo")
```
lydell, line 75: elm-review is flagging `n` as unused in this and the following tests.

- Should it be used in an expectation?
- Are `"boo"` and `"x"` the best strings to fail with?

Janiczek: I'll make the strings a bit more meaningful :+1: and change `\n` into `\_`

elm/elm.json
lydell: I am now very confused what this `elm/elm.json` file is for.

When running tests, we generate an Elm application deep in `elm-stuff/` with source-directories that include:

- Some generated Elm code in `elm-stuff/`.
- The `elm/` folder in this repo (it is part of the npm package).
- Folders from the user’s project.

`elm/elm.json` is explicitly included in the npm package: https://github.com/rtfeldman/node-test-runner/blob/026581eb144a32defd6dae3c3de5c46e5e987527/package.json#L32. But I wonder if that is a mistake.

My current thinking is that `elm/elm.json` only exists to:

1. Make it easier to work on the Elm code in `elm/` – with an elm.json editor integrations understand the code.
2. Run the unit tests of `elm/`.

But when actually running `elm-test`, I think we add the same dependencies as in `elm/elm.json` in this solver code: https://github.com/rtfeldman/node-test-runner/blob/026581eb144a32defd6dae3c3de5c46e5e987527/lib/Solve.js#L61-L64.

I think we should:

- Check if my understanding of `elm/elm.json` is correct, and if so continue with the following points.
- I guess it doesn’t matter super much which dependencies are in `elm/elm.json` as long as elm-review and the tests are happy.
- Make sure that the solver code adds all the necessary dependencies for 2.0. Is `turboMaCk/queue` needed now? I’m a bit confused.
- Add a code comment in the solver code explaining what needs to be in sync.
- Remove `elm/elm.json` from package.json.

Janiczek: `turboMaCk/queue` is needed for elm-explorations/test master / future 2.0.0

lydell: After testing, my understanding seems correct. I removed `node_modules/elm-test/elm/elm.json` in a project and it still worked. So it’s not needed.

I went ahead and cleaned up `elm/elm.json`, removed it from `package.json` and added comments in the solver code.

lydell:
TODO: Remove `turboMaCk/queue` from `elm/elm.json` when https://github.com/elm-explorations/test/pull/198 is merged

PR #616, opened 2022-10-11
Add option to clear the console

stoeffel:
We are running `elm-test  --watch` with https://github.com/DarthSim/overmind
Unfortunately, it doesn't only clear the output for elm-test, but everything else that is running through overmind.
This PR adds an option to preserve the output and printing a separate between reruns.
Let me know if this makes sense to you or if you want me to change anything. 
Thanks!

### DIFF
--- a/lib/RunTests.js
+++ b/lib/RunTests.js
@@ -43,7 +43,10 @@ function getPipeFilename(runsExecuted /*: number */) /*: string */ {
 // `elm.json changed > Compiling-- NAMING ERROR - File.elm`
 // Also note that using too much `clearLine` or `clearConsole` causes flickering
 // on Windows, so it's nicer to cleverly overwrite old output when possible.
-function makeProgressLogger(report /*: typeof Report.Report */) {
+function makeProgressLogger(
+  report /*: typeof Report.Report */,
+  preserveOutput /*: boolean */
+) {
   const items = [];
   return {
     log(message) {
@@ -67,13 +70,13 @@ function makeProgressLogger(report /*: typeof Report.Report */) {
     },
     clearLine() {
       items.length = 0;
-      if (!Report.isMachineReadable(report)) {
+      if (!Report.isMachineReadable(report) && !preserveOutput) {
         readline.clearLine(process.stdout, 0);
       }
     },
     clearConsole() {
       items.length = 0;
-      if (!Report.isMachineReadable(report)) {
+      if (!Report.isMachineReadable(report) && !preserveOutput) {
         process.stdout.write(
           process.platform === 'win32'
             ? '\x1B[2J\x1B[0f'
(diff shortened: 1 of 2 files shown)


lib/RunTests.js
```diff
@@ -67,13 +70,13 @@ function makeProgressLogger(report /*: typeof Report.Report */) {
     },
     clearLine() {
       items.length = 0;
-      if (!Report.isMachineReadable(report)) {
+      if (!Report.isMachineReadable(report) && !preserveOutput) {
```
lydell, line 73: I think we still want to allow clearing the current line?

lydell:
Here are some thoughts:

- Thanks for providing the use case! And thanks for sharing `overmind` – that’s very interesting to me personally given that I’ve written [run-pty](https://github.com/lydell/run-pty) 😄 
- [elm-test-rs](https://github.com/mpizenberg/elm-test-rs) never clears the screen. I like that the separator in elm-test-rs and this PR are very similar!
- I’d like to bikeshed the flag name later. Maybe `--no-clear-screen` is more obvious?
- We plan to release elm-test 0.19.1-revision10 tonight (https://github.com/rtfeldman/node-test-runner/pull/614). So the timing of this PR is … either perfect or unfortunate.
  - It is tempting to merge this and get it out immediately, since it’s a simple feature that makes sense. But it’s often dangerous to push things out so quickly 😅 
  - elm-test 0.19.1-revision10 is going to be compatible _only_ with elm-explorations/test 2.x. So if you’re not ready to upgrade to 2.x right now, getting this feature into that release won’t help you. We _could_ make 0.19.1-revision10 _only_ contain this PR, and then do 0.19.1-revision11 with the breaking stuff, but I’m not sure… (Also you could potentially use elm-test-rs instead until upgrading to 2.x?) 🤔

lib/RunTests.js
```diff
@@ -284,6 +289,10 @@ function runTests(
         });
         if (currentRun === undefined) {
           progressLogger.clearConsole();
+          if (preserveOutput) {
+            console.log(chalk.blue('\n\n\nRerunning tests due to changes...'));
+            console.log(chalk.blue('=================================\n\n\n'));
```
lydell, line 294: The intention is to have the `====` perfectly underline the message, right? If so, something like this might make that more clear:

```suggestion
            const message = 'Rerunning tests due to changes...';
            const separator = '='.repeat(message.length);
            console.log(chalk.blue(`\n\n\n${message}\n${separator}\n\n\n`));
```

lib/RunTests.js
stoeffel: yes, you are right

lib/RunTests.js
stoeffel: yes, makes sense

stoeffel:
@lydell Thanks for your feedback. I committed some changes to address it.
I think we are going to be able to upgrade to elm-explorations/test v2 so would be awesome to have this change in the release today, but it's not a big deal if it isn't, because as a work around we temporarily switched to elm-test-rs.

lydell:
Thanks! This is perfect, I’m adding it to the release! 🎉

PR #635, opened 2024-07-08
Fix tests due to Node.js changing the JSON.parse error messages

### LINKED ISSUE #633: Mocha tests fail on modern Node versions
It looks like the repo could use some cleanup :)
The solution is the same as https://github.com/jfmengels/node-elm-review/issues/118#issuecomment-2174238016
The root cause is that JSON.parse returns better errors on modern node versions, so my node 20 has a different error than the repo's node 14(? 16?) (12 is the engine, so maybe that?)
What node do we want for tests? Should we patch the tests to ~equal? Patch the output? How do we want to go about fixing this?
With `elm-review`, we locked our node, but then we can't test for compatibility. I'd probably patch the tests to say it should equal one of different messages.

Found testing mpizenberg/elm-solve-deps-wasm#3.

lydell:
Fixes #633.

Closes #634.

CI failed, due to the test matrix, so updated that too. Then macOS-latest required elm-tooling and elm-json to be updated.

### DIFF
--- a/tests/ElmJson.js
+++ b/tests/ElmJson.js
@@ -44,8 +44,16 @@ describe('handling invalid elm.json', () => {
           const elmJson = ElmJson.read(fullPath);
           ElmJson.requireElmTestPackage(fullPath, elmJson);
         },
-        {
-          message: expected,
+        (error) => {
+          assert.strictEqual(
+            error.message.replace(
+              // Handle slightly different JSON.parse error messages on different Node.js versions.
+              /^.+ in JSON at position .+$/gm,
+              '(the JSON parse error)'
+            ),
+            expected
+          );
+          return true;
         }
       );
     });
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,8 +12,14 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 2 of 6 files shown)


tests/ElmJson.js
```diff
@@ -44,8 +44,16 @@ describe('handling invalid elm.json', () => {
...
+        (error) => {
+          assert.strictEqual(
+            error.message.replace(
+              // Handle slightly different JSON.parse error messages on different Node.js versions.
+              /^.+ in JSON at position .+$/gm,
+              '(the JSON parse error)'
+            ),
+            expected
+          );
+          return true;
```
lishaduck, line 56: You can pass an assertion there 🤦‍♂️

lydell: Yeah, I had to look that up the docs and play around with it in my editor. Otherwise I would have left a comment on your PR. How that’s ok 😄 

This turned out to be difficult to merge anyway due to other things being out of date, and me trying to ambitious not just do the simplest fix…

lishaduck: That's fine 😁
I'd be the same way 😅

lishaduck:
🎉[^1]

[^1]: For the much felt lack of emoji reactions to merges 😉

PR #636, opened 2024-07-09
Widen constraint for `elm-solve-deps-wasm`

lishaduck:
Allow `elm-solve-deps-wasm` v2.

_**CC:** @lydell._

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,15 +1,49 @@
 {
   "name": "elm-test",
   "version": "0.19.1-revision12",
-  "lockfileVersion": 1,
+  "lockfileVersion": 3,
   "requires": true,
-  "dependencies": {
-    "@eslint/eslintrc": {
+  "packages": {
+    "": {
+      "name": "elm-test",
+      "version": "0.19.1-revision12",
+      "license": "BSD-3-Clause",
+      "dependencies": {
+        "chalk": "^4.1.2",
+        "chokidar": "^3.5.3",
+        "commander": "^9.4.1",
+        "cross-spawn": "^7.0.3",
+        "elm-solve-deps-wasm": "^1.0.2",
+        "glob": "^8.0.3",
+        "graceful-fs": "^4.2.10",
+        "split": "^1.0.1",
+        "which": "^2.0.2",
+        "xmlbuilder": "^15.1.1"
+      },
+      "bin": {
+        "elm-test": "bin/elm-test"
+      },
+      "devDependencies": {
+        "elm-review": "2.9.1",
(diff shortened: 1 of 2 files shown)


lydell:
Is this to save space in node_modules/ for those who install both elm-review and elm-test?

lishaduck:
> Is this to save space in node_modules/ for those who install both elm-review and elm-test?

Yeah. If we allow both versions, then it'll the latest.
I did the same in elm-review. This way, it's not breaking, we don't force anyone to upgrade. However, npm will default to v2.
If you install elm-test, then elm-review, currently you'll only get one, but not the other way around. This fixes that.

lishaduck:
@lydell, ping :)

lydell:
I’ve been waiting for a release of elm-review with bumps :)

lishaduck:
I think we're currently targeting a major (dropping Node 10&12), so it'll probably be a while before we release, but this breaks our CI (we depend on 2.0.0's types internally).

Alternatively, @jfmengels, would you be ok with backporting the `elm-solve-deps-wasm` upgrade to elm-review 2.x?
I think that'd just need a new branch off jfmengels/node-elm-review@5f2fcc3138b7fe01b315590e0546ce1255cf21c8, cherrypick jfmengels/node-elm-review@79e6b9ab092455f4585f0b875a4ef83ba6ae4070, and get tests to pass (i.e., bump elm-syntax in tests).

jfmengels:
@lishaduck I think we're still on course for v2 for `elm-review`. We're officially dropping support for v10/12, but they were already broken for a while (so the only difference would be the "officially" part). I think I'd be happy to release a new version. What do you think @lishaduck?

lydell:
There’s no rush with this, right? The new version does not bring any improvements to end users?

lishaduck:
> There’s no rush with this, right? The new version does not bring any improvements to end users?

`elm-solve-deps-wasm` v2 fixes incorrect types. The incorrect types breaks `elm-review`'s CI. There's not any improvements to `elm-solve-deps-wasm` for end users, which is why there isn't a new `elm-review` bump planned. Because we depend on `elm-test`, though, npm prefers installing the old version. It's not terribly important (I can manually fix the lockfile, and npm'll go with it), but it'd be nice.

lishaduck:
> I’ve been waiting for a release of elm-review with bumps :)

New elm-review out with bumps :)

(Again, not a big deal, but it is a nice QoL improvement)

lydell:
Released in 0.19.1-revision13. Thanks for the reminder!

Issue #646, opened 2025-02-15
Consider switching to tinyglobby


lishaduck:
It supports down through node12, is faster (allegedly), has a nicer API (subjectively), and is what elm-review uses nowadays.
Oh, and b/c npm/npm-pick-manifest#33 didn't land until npm 10.8.2 (node 22.5.0), you'll ideally not want to depend on multiple versions of glob (elm-review hit this), so tinyglobby also helps there (meaning the workarounds from #639 are unneeded).

lydell:
Oh, thanks for the heads up! So elm-test 0.19.1-revision13 is broken for Node.js 10–18 now (using the default npm that ships with those Node.js versions), fml 😵

lydell:
I did a quick-fix for now, restoring Node.js 14–18 support. Node.js 12 is still broken.

Switching to tinyglobby sounds great, but isn’t something I want to spend time on right now.

jfmengels:
From what I can see, it's broken (at least) because of `minimatch` v9 being used, which only supports 14 and up. `minimatch` is used because `glob@10`, so switching to `tinyglobby` (v12 and up) might be a good solution to the problem actually.

PR #647, opened 2025-02-15
Replace glob by tinyglobby

jfmengels:
This should I think fix the support for Node.js v12 (and fixes https://github.com/rtfeldman/node-test-runner/issues/646)

> From what I can see, Node.js v12 broken (at least) because of minimatch v9 being used, which only supports 14 and up. minimatch is used because glob@10, so switching to tinyglobby (v12 and up) might be a good solution to the problem actually.
(https://github.com/rtfeldman/node-test-runner/issues/646#issuecomment-2660864905)

Note: I haven't tested it as thoroughly as I'd like to yet, but it seems to work well so far.

The packages are relatively interchangeable, though the options change; Here is the documentation for each:
- tinyglobby: https://www.npmjs.com/package/tinyglobby#Options
- glob: https://www.npmjs.com/package/glob#Options

`nodir` becomes `onlyFiles: true`
`nocase` becomes `caseSensitiveMatch` (but inverted)
I could not find an equivalent to glob's `mark` (Append a / on any directories matched) but that is the default behavior for `tinyglobby` anyway.

From what I can tell, these are the paths through which globbing is used (My understanding of what is happening at least):
In `resolveGlobs`, we look for test files. By default we look for `<root>/tests`, but that can be overridden by CLI arguments (`elm-test otherDir/ tests/SpecificTest.elm`).We first iterate through that list to understand which each of these paths exist or not. If it exists, we defer to `findAllElmFilesInDir` if it's a directory, and just take the file it it's a file. If it does not exist, then it's likely a glob (`elm-test "**/*Test.elm" "te*ts/"`, which we defer to `resolveCliArgGlob` where we apply the glob. For every directory, we then again defer to findAllElmFilesInDir`.

In findAllElmFilesInDir`, we use globbing to find all contained Elm files.

### DIFF
--- a/lib/FindTests.js
+++ b/lib/FindTests.js
@@ -71,7 +71,7 @@ function resolveCliArgGlob(
   // https://github.com/isaacs/minimatch/commit/9104d8d175bdd8843338103be1401f80774d2a10#diff-f41746899d033115e03bebe4fbde76acf2de4bf261bfb221744808f4c8a286cf
   const pattern =
     process.platform === 'win32'
-      ? globRelativeToProjectRoot.replace(/\\/g, '/')
+      ? globRelativeToProjectRoot //.replace(/\\/g, '/')
       : globRelativeToProjectRoot;
 
   return globSync(pattern, {
--- a/lib/FindTests.js
+++ b/lib/FindTests.js
@@ -59,19 +59,16 @@ function resolveCliArgGlob(
     path.resolve(fileGlob)
   );
 
-  // glob@8 (via minimatch@5) had a breaking change where you _have_ to use
-  // forwards slash as path separator, regardless of platform, making it
-  // unambiguous which characters are separators and which are escapes. This
-  // restores the previous behavior, avoiding a breaking change in elm-test.
+  // The globs _have_ to use forwards slash as path separator, regardless of
+  // platform, making it unambiguous which characters are separators and which
+  // are escapes.
   // Note: As far I can tell, escaping glob syntax has _never_ worked on
   // Windows. In Elm, needing to escape glob syntax should be very rare, since
   // Elm file paths must match the module name (letters only). So it’s probably
   // more worth supporting `some\folder\*Test.elm` rather than escaping.
-  // https://github.com/isaacs/node-glob/issues/468
-  // https://github.com/isaacs/minimatch/commit/9104d8d175bdd8843338103be1401f80774d2a10#diff-f41746899d033115e03bebe4fbde76acf2de4bf261bfb221744808f4c8a286cf
   const pattern =
     process.platform === 'win32'
-      ? globRelativeToProjectRoot //.replace(/\\/g, '/')
+      ? globRelativeToProjectRoot.replace(/\\/g, '/')
       : globRelativeToProjectRoot;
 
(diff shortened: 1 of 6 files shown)


lib/FindTests.js
```diff
@@ -74,29 +74,27 @@ function resolveCliArgGlob(
...
 function findAllElmFilesInDir(dir /*: string */) /*: Array<string> */ {
-  return glob.sync('**/*.elm', {
+  return globSync('**/*.elm', {
     cwd: dir,
-    nocase: true,
+    caseSensitiveMatch: false,
     absolute: true,
     ignore: ignoredDirsGlobs,
-    nodir: true,
+    onlyFiles: true,
```
jfmengels, line 97: `onlyFiles: true` is the default, but I've kept it for explicitness.

lib/FindTests.js
```diff
@@ -74,29 +74,27 @@ function resolveCliArgGlob(
...
-      absolute: true,
-      ignore: ignoredDirsGlobs,
-      // Match directories as well and mark them with a trailing slash.
-      nodir: false,
-      mark: true,
-    })
-    .flatMap((filePath) =>
-      filePath.endsWith('/') ? findAllElmFilesInDir(filePath) : filePath
-    );
+  return globSync(pattern, {
```
lydell, line 77: Note to self: there a big comment above about glob@8 and Windows and backslashes. Try it out on Windows to see how it behaves in tinyglobby, and remove or update the code/comment?

lib/FindTests.js
```diff
@@ -74,29 +74,27 @@ function resolveCliArgGlob(
...
+    cwd: projectRootDir,
+    caseSensitiveMatch: false,
+    absolute: true,
+    ignore: ignoredDirsGlobs,
+    // Match directories as well
+    onlyFiles: false,
+  }).flatMap((filePath) =>
+    // Directories have their path end with `/`
+    filePath.endsWith('/') ? findAllElmFilesInDir(filePath) : filePath
```
lydell, line 86: Note to self: Does it end with a slash on Windows? We don’t seem to have test coverage for this condition being true.

lydell (commented):
Thank you so much! ⭐ 

I think I have to do some manual tests on Windows before I dare merging this.

lib/FindTests.js
lydell: Added test case: [`36cdb47` (#647)](https://github.com/rtfeldman/node-test-runner/pull/647/commits/36cdb477d77236767181d7c4bb36086687defe19)

Btw, that test was failing on master – see https://github.com/rtfeldman/node-test-runner/pull/648. But it passed when installing glob 8 instead of 10. Another breakage by glob 🙈

lib/FindTests.js
lydell: It’s needed: [`7814ff7` (#647)](https://github.com/rtfeldman/node-test-runner/pull/647/commits/7814ff7badeaf58024a150c5b045dfc259b47373)

Updated the comment: [`46ca4aa` (#647)](https://github.com/rtfeldman/node-test-runner/pull/647/commits/46ca4aa27f4104ac1d7e34acecb4304939e3df69)

PR #650, opened 2025-07-06
Make compatible with upcoming version of Lamdera

miniBill:


lydell:
Awesome!

Having read https://github.com/lamdera/compiler/issues/56 before, I expected the diff to be something like `F2\(` turned into `(F2\()?` – but nope, it’s an added `\$?`! So Lamdera is now adding an extra `$` to variable names? Just so that I understand the whole thing, which change in Lamdera is that from? And out of curiosity – why?

miniBill:
The change is from the Fx/Ax patch. The added $ is the "unwrapped" version, whereas the version without is the original Fx-wrapped one.

PR #659, opened 2026-02-20
Vendor chalk and supports-color

jfmengels:
Vendors the `chalk` dependency, getting rid of a few indirect dependencies too.
`chalk` does quite a lot of complicated work to enable complex usages (`chalk.blue.underline(...)`, template literals, ...), whereas we only use it to do `chalk.red()` and `chalk.blue()` (and that's it).

We also `chalk` it to figure out whether the output supports colors, but that is handled by the sub-dependency `supports-color`.

I've vendored both dependencies, and simplified them. `chalk` is now extremely simple, and `supports-color` is simplified a bit (I've done it in steps for it to be understandable, or even replicable later). Simplifying these dependencies mean startup should be faster :crossed_fingers: 

I'm still on the fence for vendoring `supports-color` (but leaning towards doing it) as we will now receive new updates for it. That said, the latest version is already incompatible with this CLI, as it's ESM only. Therefore we won't get any updates until we switch to ESM (thereby dropping Node.js v12 compatibility).


lydell (commented):
Great! Pretty funny that we depended on a whole library plus sub dependencies just to make some text blue and red 😅

jfmengels:
I like the changes :+1:

PR #671, opened 2026-07-19
Remove cross-spawn dependency

lydell:


tests/ci.js
```diff
@@ -23,9 +23,9 @@ const resultFailureThreshold = 2;
  * @returns { import('child_process').SpawnSyncReturns<string> }
  */
 function execElmTest(args, cwd = '.') {
-  return spawn.sync(
-    elmtestPath,
-    args,
+  return Spawn.spawnSync(
+    'node',
+    [elmTestPath, ...args],
```
lydell, line 28: cross-spawn supports parsing shebangs and making them work on Windows. Our own replacement does not (because it’s not needed when executing `elm`). To make the tests work, we simply call `node` explicitly instead of relying on the shebang.

PR #680, opened 2026-07-21
Make --workers=1 single-threaded

lydell:
This is slightly faster, and being able to run everything in the same process makes debugging easier.

View the [diff without whitespace changes](https://github.com/rtfeldman/node-test-runner/pull/680/changes?w=1) since a lot of code became indented.

### DIFF
--- a/elm/src/Test/Runner/Node.elm
+++ b/elm/src/Test/Runner/Node.elm
@@ -81,7 +81,7 @@ type Msg
 {-| The port names are prefixed to reduce the likelihood of the project
 having a port with the same name, which is a compile error.
 -}
-port elmTestPort__send : String -> Cmd msg
+port elmTestPort__send : Decode.Value -> Cmd msg
 
 
 port elmTestPort__receive : (Decode.Value -> msg) -> Sub msg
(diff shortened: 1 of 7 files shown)


lib/Generate.js
```diff
@@ -68,6 +72,8 @@ function addKernelTestChecking(content) {
         checkDefinition,
         '$1 = value => value && value.__elmTestSymbol === __elmTestSymbol ? $elm$core$Maybe$Just(value) : $elm$core$Maybe$Nothing;'
       )
+      // Simply remove the first occurrence of `console.warn`. This leaves the message string in parentheses behind, but that’s fine.
+      .replace('console.warn', '')
```
lydell, line 76: This replaces the previous approach of re-assigning `console.warn`. This is simpler, and avoids messing with the global `console.warn` in the main thread.

elm/src/Test/Runner/Node.elm
```diff
@@ -81,7 +81,7 @@ type Msg
 {-| The port names are prefixed to reduce the likelihood of the project
 having a port with the same name, which is a compile error.
 -}
-port elmTestPort__send : String -> Cmd msg
+port elmTestPort__send : Decode.Value -> Cmd msg
```
lydell, line 84: We now turn this into a JSON string in JS instead – and not at all in single-threaded mode.

PR #684, opened 2026-08-09
Add --dependencies and --offline options

lydell:
This adds `--dependencies` and `--offline` options, just like `elm-test-rs`.

`--dependencies oldest` is useful for package development. This is implemented by reversing the list of available versions for a package.

I added `--offline` in the same PR, because [`elm-test-rs` documents that `--offline` does not work with `--dependencies`](https://github.com/mpizenberg/elm-test-rs/tree/a0345e60af6248b976bcb600c9ebd26b83edce0b#offline-mode) and that made me think more about online and offline and what “newest” and “oldest” actually mean. To have `--dependencies oldest` and `--dependencies newest` work correctly and in non-surprising ways, I needed to rethink some caching and online/offline behaviors around dependency solving.

I also noted that the error message when a solve failed looked a bit out of context, and that we showed no progress output while solving online (slow). Before:

```

Because there is no version of elm/random in 1.0.0 <= v < 2.0.0 and elm-explorations/node-test-runner-example-package 1.0.0 depends on elm/random 1.0.0 <= v < 2.0.0, elm-explorations/node-test-runner-example-package 1.0.0 is forbidden.
```

After:

```
Solving dependencies
Failed to solve dependencies:
Because there is no version of elm/random in 1.0.0 <= v < 2.0.0 and elm-explorations/node-test-runner-example-package 1.0.0 depends on elm/random 1.0.0 <= v < 2.0.0, elm-explorations/node-test-runner-example-package 1.0.0 is forbidden.
```

### DIFF
--- a/lib/DependencyProvider.js
+++ b/lib/DependencyProvider.js
@@ -241,7 +241,7 @@ class DependencyProvider {
           lister.list(pkg, toOrderConstraints(elmJson, packageStrategy, pkg))
       );
     } catch (errorMessage) {
-      throw new Error(errorMessage);
+      throw new Error(`Failed to solve dependencies:\n${errorMessage}`);
     }
   }
 
@@ -269,7 +269,7 @@ class DependencyProvider {
           lister.list(pkg, toOrderConstraints(elmJson, packageStrategy, pkg))
       );
     } catch (errorMessage) {
-      throw new Error(errorMessage);
+      throw new Error(`Failed to solve dependencies:\n${errorMessage}`);
     }
   }
 }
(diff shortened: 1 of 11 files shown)


lib/DependencyProvider.js
```diff
@@ -119,7 +119,7 @@ class OnlineVersionsCache {
   }
 }
 
-class OnlineAvailableVersionLister {
+class AvailableVersionLister {
```
lydell, line 122: It turned out that the `OnlineAvailableVersionLister` wasn’t doing anything online, and the `OfflineAvaileableVersionLister` was doing the same work, but wrongly only looked in one place (`~/.elm/0.19.2/packages/`, but not `~/.elm/pubgrub/`). So I removed the “offline” one and renamed the “online” one.

PR #686, opened 2026-08-16
Use `Test.RunnerV2`

lydell:
Use `Test.RunnerV2` from https://github.com/elm-explorations/test/pull/260, and implement everything that unlocks.

More info coming soon.


elm/src/Test/Reporter/Console.elm
```diff
@@ -36,6 +35,22 @@ pluralize singular plural count =
...
+    -> List Text
+formatLabels formatDescription formatTest labels =
+    case labels of
+        [] ->
+            []
+
+        test :: descriptions ->
+            formatTest test
+                :: List.map formatDescription descriptions
+                |> List.reverse
```
jfmengels: For a small speedup boost, you can inline `List.map`:

```elm
formatTest test
  :: List.foldr (\x acc -> formatDescription x :: acc) [] descriptions
  |> List.reverse
```

then move `formatTest test` to the initial accumulator and use `List.foldl` instead of `List.foldr`:

```elm
List.foldl
  (\x acc -> formatDescription x :: acc)
  [ formatTest test ]
  descriptions
```

elm/src/Test/Runner/Ports.elm
```diff
@@ -0,0 +1,150 @@
...
+        (Encode.object
+            [ ( "type", Encode.string "READY" )
+            , ( "unitTests", Encode.list Encode.int unitTests )
+            , ( "fuzzTests", Encode.list Encode.int fuzzTests )
+            ]
+        )
+
+
+sendResult : Int -> Bool -> String -> List String -> Maybe String -> Decode.Value -> Decode.Value -> Cmd msg
+sendResult testId isFuzzTest jsDefinitionName labels expectationElmCode debugLogs report =
```
jfmengels: Maaaybe worth using a record to make function calls easier to read?

elm/src/Test/Runner/Node.elm
```diff
@@ -8,107 +8,376 @@ port module Test.Runner.Node exposing (check, run, TestProgram)
...
+    | CachedUnitTestFail
+        { description : String
+        , reason : Reason
+        }
+
+
+{-| Non-opaque version of `FuzzTestExpectation`, but without `rerunFailure`.
+-}
+type CachedFuzzTestExpectation
+    = CachedFuzzTestPass
```
jfmengels: Could be turned into `CachedFuzzTestPass DistributionReport`

elm/src/Test/Runner/Node.elm
```diff
@@ -8,107 +8,376 @@ port module Test.Runner.Node exposing (check, run, TestProgram)
...
-            sendResults True model.testReporter model.results
+            Ports.sendError ("Unit test not found: " ++ String.fromInt testId)
 
-        Just config ->
+        Just unitTest ->
             let
-                outcome =
-                    outcomeFromExpectations (config.run ())
+                -- The unnecessary-looking tuple here ensures that `getAndClearDebugLogs`
+                -- runs _after_ the thunk.
```
jfmengels: You could also do
```elm
let
  ( expectation, duration ) =
    runWithDuration (\() -> Runner.runUnitTest unitTest)
  -- ...
in
sendUnitTestResult testId unitTest cachedExpectation duration (getAndClearDebugLogs False) model.testReporter
```

or
```elm
let
  ( expectation, duration ) =
    runWithDuration (\() -> Runner.runUnitTest unitTest)
in
let
  debugLogs =
    getAndClearDebugLogs False
  -- ...
in
-- ...
```
In both cases, this is a clearer sign that `getAndClearDebugLogs` will be called after `runWithDuration`. The order of computation of a tuple seems too incidental to rely on comfortably I'd say. (Also, one less tuple allocation)

I think a comment indicating that `getAndClearDebugLogs` has to be computed afterwards because it relies on side-effects from `runWithDuration` would be welcome still.

elm/src/Test/Runner/Node.elm
```diff
@@ -8,107 +8,376 @@ port module Test.Runner.Node exposing (check, run, TestProgram)
...
+                ( ( expectation, duration ), debugLogs ) =
+                    ( runWithDuration (\() -> Runner.runUnitTest unitTest), getAndClearDebugLogs False )
+
+                cachedExpectation =
+                    case expectation of
+                        UnitTestPass ->
+                            CachedUnitTestPass
```
jfmengels: I feel like the conversion a custom type is unnecessary, especially as it gets reconverted to something else right away in `sendUnitTestResult`

elm/src/Test/Runner/Node.elm
```diff
@@ -8,107 +8,376 @@ port module Test.Runner.Node exposing (check, run, TestProgram)
...
+        hasDebugLogs =
+            not (isEmptyDebugLogs debugLogs)
+
+        outcome =
+            case expectation of
+                CachedUnitTestPass ->
+                    Passed NoDistribution
+
+                CachedUnitTestFail { description, reason } ->
+                    if reason == TODO then
```
jfmengels: Use a case expression maybe? I think that makes for a faster check in the compiled code.

elm/src/Test/Runner/Node.elm
```diff
@@ -8,107 +8,376 @@ port module Test.Runner.Node exposing (check, run, TestProgram)
...
+
+        report =
+            testReporter.reportComplete result
+
+        expectationElmCode =
+            if expectation == CachedUnitTestPass && not hasDebugLogs then
+                Nothing
+
+            else
+                Just (Debug.toString expectation)
```
jfmengels, line 276: What is the output of this? Would it start with `CachedUnitTestFail ...`? If so, that feels a bit odd.

lib/Generate.js
```diff
@@ -74,6 +130,58 @@ function patch(content) {
...
+  } else {
+    _Debug_logs.push(tag + ': ' + _Debug_toString(value));
+  }
+  return value;
+});
+            `.trim()
+      )
+      .replace(
+        getAndClearDebugLogsDefinition,
+        '$1 = paused => { var logs = _Json_wrap(_Debug_logs); _Debug_logs = []; _Debug_logPaused = paused; return logs; }'
```
jfmengels: Use `function(paused)` instead of `=>` to keep compiling to ES5?

elm/src/Test/Runner/Node.elm
```diff
@@ -8,107 +8,376 @@ port module Test.Runner.Node exposing (check, run, TestProgram)
...
+                    else
+                        []
+
+                seed =
+                    Random.initialSeed model.runInfo.initialSeed
+
+                ( expectation, duration, debugLogs ) =
+                    -- Pause debug logs.
+                    getAndClearDebugLogs True
+                        |> (\_ ->
```
jfmengels: Seems like there's a number of places where we ignre the returned logs. Should there be a variant of `getAndClearDebugLogs` that simply clears `Debug.log`s?

lib/Hash.js
```diff
@@ -0,0 +1,226 @@
...
+}
+
+/**
+ * Merges recursive chains into single items, so that the output is an acyclic graph.
+ *
+ * @param { Record<string, Node> } graph
+ * @returns { void }
+ */
+function makeAcyclicStep(graph) {
+  const scc = Tarjan.stronglyConnectedComponents({
```
jfmengels, line 140: Do you know if these chains are reduced to their smallest possible cycles? (Similar to the Elm import cycle sometimes including unrelated modules)

README.md
```diff
@@ -156,12 +156,30 @@ Start the runner in watch mode. Your tests will automatically rerun whenever you
...
+
+By default, the console is cleared before each run in watch mode, so you only see the latest information. If you don’t like this, turn it off with `--no-clear-console`.
+
+    elm-test --watch --no-clear-console
+
+### --unbuffered-logs
+
+elm-test collects all `Debug.log` output while executing a test, and displays it all once the test in question is finished. This way elm-test can print _which_ test the logs came from.
+
+If the function your are testing gets into an infinite loop, it means that your debug logs will never show up. Then it can be useful to have the logs print _immediately_ instead (at the loss of no longer being able to label which tests the logs came from). To avoid confusion, use [Test.only](https://package.elm-lang.org/packages/elm-explorations/test/latest/Test#only) to isolate your test, or pass `--workers 1` to run in single-threaded mode to avoid oddly mixed output:
```
jfmengels:
```suggestion
If the function you are testing gets into an infinite loop, it means that your debug logs will never show up. Then it can be useful to have the logs print _immediately_ instead (at the loss of no longer being able to label which tests the logs came from). To avoid confusion, use [Test.only](https://package.elm-lang.org/packages/elm-explorations/test/latest/Test#only) to isolate your test, or pass `--workers 1` to run in single-threaded mode to avoid oddly mixed output:
```

jfmengels (commented):
Great work I'm not sure I reviewed everything, it was a lot of changes, but here are a few suggestions I have.

I think at this point it would be better for me to review by trying it out. If you have some instructions on how to use the 2 branches together, I would love to see them.

elm/src/Test/Runner/Ports.elm
lydell: The call sites look like this:

```elm
Ports.sendResult testId False jsDefinitionName labels expectationElmCode debugLogs report
```

```elm
Ports.sendResult testId True jsDefinitionName labels expectationElmCode debugLogs report
```

So a record would mostly be 🦜 (but give a name to the boolean).

I wonder what’s faster, passing a big record or many parameters. My guess would be many parameters (up to a point).

elm/src/Test/Runner/Node.elm
lydell: It is this way because `sendUnitTestResult` is also called in a different place – where we read test expectations from cache instead of running tests. Also, we need to construct a `CachedUnitTestExpectation` so that we can `Debug.toString` it for caching.

elm/src/Test/Runner/Node.elm
lydell: Yes, for example:

```elm
CachedUnitTestFail { description = "Expect.equal", reason = Equality "\" blah\"" "\"blah\"" }
```

It can also be `CachedUnitTestPass` if there are debug logs.

Why does it feel a bit odd?

lib/Hash.js
lydell: I don’t. All I know is that I naively “solved” recursive function chains myself, and kept finding cases where it broke. Then I read up on it and learned that it is a well-known computer science graph problem, and that Tarjan’s algorithm can solve it. Then I found an implementation of it on npm with _lots_ of dependents that looked good. Since I started using it I haven’t run into issues.

lydell:
> If you have some instructions on how to use the 2 branches together, I would love to see them.

In elm-explorations/test, run `node --run install-local && rm -f ~/.elm/0.19.2/packages/registry.dat` to copy the local version of the package into `~/.elm/`.

In node-test-runner:

1. `cd example-application`
2. `rm -rf elm-stuff`
3. `../bin/elm-test`

To test it on some other project’s test suite:

1. Make sure that the project uses Elm 0.19.2.
2. Make sure that the project uses `"elm-explorations/test": "2.2.1"`.
3. `rm -rf elm-stuff`
4. `~/path/to/node-test-runner/bin/elm-test`

elm/src/Test/Runner/Ports.elm
jfmengels: Alright that's fair.

> I wonder what’s faster, passing a big record or many parameters.

Haven't checked it, but I'm pretty sure many parameters is faster. (will try to add a benchmark for that in my benchmark repo :eyes: )

elm/src/Test/Runner/Node.elm
lydell: Maybe I should mention that it is used for codegen:ing the cache file.

elm/src/Test/Runner/Node.elm
lydell: This code does not exist anymore. Debug logs collection and test timing is not done in elm-explorations/test (in a much cleaner way).

elm/src/Test/Runner/Node.elm
lydell: This code does not exist anymore. Debug logs collection and test timing is not done in elm-explorations/test (in a much cleaner way).

elm/src/Test/Runner/Ports.elm
lydell: I have pushed more updates, and some call sites are definitely a bit cryptic now. Might be worth using a record.

lib/Generate.js
lydell: I have gotten rid of all patching of the JS output now.

elm/src/Test/Runner/Node.elm
```diff
@@ -135,272 +462,358 @@ update msg ({ testReporter } as model) =
...
-                }
-            , processes = 0
-            , nextTestToRun = 0
-            , results = []
-            , testReporter = createReporter report
-            , autoFail = Nothing
-            }
+trawlNext : Cmd Msg
+trawlNext =
+    Task.perform (\() -> Trawl) (Task.succeed ())
```
jfmengels, line 767: I believe this also works
```suggestion
    Task.perform identity Trawl
```

lydell:
```
-- TYPE MISMATCH -------------------------------------- src/Test/Runner/Node.elm

The 2nd argument to `perform` is not what I expect:

767|     Task.perform identity Trawl
                               ^^^^^
This `Trawl` value is a:

    Msg

But `perform` needs the 2nd argument to be:

    Task.Task Never msg

Hint: I always figure out the argument types from left to right. If an argument
is acceptable, I assume it is “correct” and move on. So the problem may actually
be in one of the previous arguments!
```

